PR #133, opened 2016-11-02
MLIBZ-1398: bugfix inside KCSHttpRequest and NSNull objects being converted to nil

heyzooi:


KinveyKit/KinveyKit/Source/KCSNetworkResponse.m
```diff
@@ -129,7 +129,7 @@ - (id) jsonResponseValue:(NSError**) anError
                                                                         options:NSJSONReadingMutableContainers
                                                                           error:&error];
     NSObject* jsonObj = nil;
-    if (![jsonResponse isKindOfClass:[NSDictionary class]]) {
+    if (!([jsonResponse isKindOfClass:[NSDictionary class]] || [jsonResponse isKindOfClass:[NSArray class]])) {
```
tejasranade: @heyzooi the variable `jsonResponse` is of `NSMutableDictionary`. In what case will it be an array?

heyzooi: if you are performing a find, the results is an array of objects

tejasranade:
Reviewed with @heyzooi . LGTM.

PR #137, opened 2016-11-09
MLIBZ-1393: be able to me provide custom fields while user signup

heyzooi:


Kinvey/KinveyTests/KinveyTestCase.swift
```diff
@@ -100,9 +156,9 @@ class KinveyTestCase: XCTestCase {
         }
         
         if let username = username {
-            User.signup(username: username, completionHandler: handler)
+            User.signup(username: username, user: user, completionHandler: handler)
```
tejasranade: @heyzooi what's the type of `user` here? You should add at least one test where `user` is a custom subclass of `User`.

Kinvey/Kinvey/User.swift
```diff
@@ -42,15 +49,15 @@ open class User: NSObject, Credential, Mappable {
...
 
-        let request = client.networkRequestFactory.buildUserSignUp(username: username, password: password)
-        Promise<User> { fulfill, reject in
+        let request = client.networkRequestFactory.buildUserSignUp(username: username, password: password, user: user)
+        Promise<UserType> { fulfill, reject in
             request.execute() { (data, response, error) in
                 if let response = response , response.isOK {
                     client.activeUser = client.responseParser.parseUser(data)
-                    fulfill(client.activeUser!)
+                    fulfill(client.activeUser as! UserType)
```
tejasranade: This would require that `client.activeUser` returns a `UserType` instance. Is that true today? Last I checked, that wasn't the case.

tejasranade:
@heyzooi  this should be accompanied by a devcenter PR for the updated code snippets. Can you make one please?

Kinvey/KinveyTests/KinveyTestCase.swift
heyzooi: the method is a generic method `signUp<UserType: User>(user: UserType?)` so it can be actually any user type

Kinvey/Kinvey/User.swift
heyzooi: same thing here, generic method `signup<UserType: User>(user: UserType?)` and the `parseUser()` method also considers the `client.userType` property

heyzooi:
Sure!

```
client.userType = MyUser.self
        
let user = MyUser()
user.foo = "bar"
User.signup(user: user) { user, error in
    if let user = user {
        //success
    } else {
        //failure
    }
}
```

PR #138, opened 2016-11-09
MLIBZ-1348: SSO implementation

heyzooi:


heyzooi:
don't merge yet, this was just the first steps that we need to do

Kinvey/Kinvey/Client.swift
```diff
@@ -175,14 +180,21 @@ open class Client: NSObject, NSCoding, Credential {
     }
     
     /// Initialize a `Client` instance with all the needed parameters and requires a boolean to encrypt or not any store created using this client instance.
-    open func initialize(appKey: String, appSecret: String, apiHostName: URL = Client.defaultApiHostName, authHostName: URL = Client.defaultAuthHostName, encrypted: Bool, schemaVersion: CUnsignedLongLong = 0, migrationHandler: Migration.MigrationHandler? = nil) {
+    @available(*, deprecated: 3.3.3, message: "Please use initialize(appKey:appSecret:accessGroup:apiHostName:authHostName:encrypted:schema:completionHandler:)")
```
tejasranade: I'm trying to understand this - if someone is not interested in SSO, do we still want them to use the longform `initialize`?

Kinvey/Kinvey/User.swift
```diff
@@ -538,7 +551,7 @@ open class User: NSObject, Credential, Mappable {
                 micLoginVC.setValue(forceUIWebView, forKey: "forceUIWebView")
             }
             micLoginVC.client = client
-            micLoginVC.micApiVersion = client.micApiVersion
+            micLoginVC.micApiVersion = client.micApiVersion?.rawValue
```
tejasranade: My preference is to use the `rawValue` only when we build the URL. In the rest of the SDK, why don't we pass the enumeration value?

Kinvey/Kinvey/Client.swift
heyzooi: so, because we now have an async handler that returns a possible user (using SSO), I thought that would better to concise the schema migration in a tuple with the scheme version and the migration handler

`public typealias Schema = (version: CUnsignedLongLong, migrationHandler: Migration.MigrationHandler?)`

Kinvey/Kinvey/User.swift
heyzooi: micLoginVC is a Obj-C class, which we plan to replace soon for a Swift version of it. So, for this specific PR the rawValue still the best way to pass through.

PR #141, opened 2016-11-16
MLIBZ-1395: adding http response and data for some error objects

heyzooi:


Kinvey/Kinvey/Error.swift
```diff
@@ -21,22 +21,22 @@ public enum Error: Swift.Error {
     case objectIdMissing
     
     /// Error when a method is not allowed, usually when you are using a Data Link Connector (DLC).
-    case methodNotAllowed(debug: String, description: String)
+    case methodNotAllowed(httpResponse: HTTPURLResponse?, data: Data?, debug: String, description: String)
```
tejasranade: @heyzooi `httpResponse` makes sense here. What does `data` represent? How is the developer supposed to use it?

Kinvey/Kinvey/Error.swift
```diff
@@ -57,44 +57,95 @@ public enum Error: Swift.Error {
...
+    
+    /// Response Header `X-Kinvey-Request-Id`
+    public var requestId: String? {
+        return httpResponse?.allHeaderFields[RequestIdHeaderKey] as? String
+    }
+    
+    /// Response Data Body object.
+    public var responseDataBody: Data? {
+        switch self {
+        case .unknownError(_, let data, _):
```
tejasranade: I prefer we squash these `case` statements to make the code more concise.

Kinvey/Kinvey/Error.swift
heyzooi: it would allow the app to get the response body if a DLC for example returns some useful information in the response body

tejasranade:
Additional changes - 
- Expose `debug` and `description` fields in the error objects to match the fields that the backend responds with.

PR #148, opened 2016-11-19
MLIBZ-1393: be able to me provide custom fields while user signup

heyzooi:


heyzooi:
Got approval already, so merging now.

PR #149, opened 2016-11-29
Remove temporary `_id` when saving a new entity

thediycoder:
### Description
If the `_id` for the entity being saved is nil or starts with `'tmp_'` then remove the `_id` from the `bodyObject`.

### Changes
- Check if `objId` starts with `ObjectIdTmpPrefix`
- set `bodyObject[PersistableIdKey]` to `nil` if `isNewObj` is `true`


thediycoder:
@heyzooi Will this cleanup the sync table properly and merge the two related entities in realm?

PR #150, opened 2016-11-29
MLIBZ-1398: returning the default value for a BOOL type when the response returns…

heyzooi:
… a nil value

### DIFF
--- a/KinveyKit/KinveyKit/Source/KCSValueConverter.m
+++ b/KinveyKit/KinveyKit/Source/KCSValueConverter.m
@@ -64,7 +64,13 @@
             }
             value = results;
         }
-        return value != [NSNull null] ? value : nil;
+        if (value != [NSNull null]) {
+            return value;
+        } else if ([valueType isEqualToString:@"B"]) {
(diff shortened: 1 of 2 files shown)


KinveyKit/KinveyKit/Source/KCSValueConverter.m
```diff
@@ -64,7 +64,13 @@ +(id)convert:(id)value
             }
             value = results;
         }
-        return value != [NSNull null] ? value : nil;
+        if (value != [NSNull null]) {
+            return value;
+        } else if ([valueType isEqualToString:@"B"]) {
+            //if the value is a NSNull and the valueType is a BOOL, represented by 'B' according to Apple's documentation: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html
+            return @NO;
+        }
```
tejasranade: @heyzooi it will help to add a catch-all here that tells the user why we couldn't process their type. Something like - 
"Your model class is trying to map a property with a `null` value to a `<valueType>` primitive. This is not supported. To fix this issue, use a `Foundation` type instead of the primitive type."

PR #152, opened 2016-11-30
Deprecation of extraneous "reset password" methods

tejasranade:
The iOS SDK has "resetPassword" in several flavors, which seems unnecessary. The class method that takes usernameOrEmail seems adequate. 
I've left the instance method as well, just because it provides a shorthand way to reset the current user's password. Other methods are now deprecated.

### DIFF
--- a/Kinvey/Kinvey/User.swift
+++ b/Kinvey/Kinvey/User.swift
@@ -202,7 +202,8 @@ open class User: NSObject, Credential, Mappable {
         return User.sendEmailConfirmation(forUsername: username, client: client, completionHandler: completionHandler)
     }
     
-    fileprivate class func resetPassword(usernameOrEmail: String, client: Client = Kinvey.sharedClient, completionHandler: VoidHandler? = nil) -> Request {
+    @discardableResult
+    open class func resetPassword(usernameOrEmail: String, client: Client = Kinvey.sharedClient, completionHandler: VoidHandler? = nil) -> Request {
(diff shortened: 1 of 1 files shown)


Kinvey/Kinvey/User.swift
```diff
@@ -222,12 +224,14 @@ open class User: NSObject, Credential, Mappable {
     
     /// Sends an email to the user with a link to reset the password using the `username` property.
     @discardableResult
+    @available(*, deprecated: 3.3.3, message: "Use resetPassword(usernameOrEmail:) instead")
```
heyzooi: 3.3.3 is the latest release. the next version would be 3.3.4

Kinvey/Kinvey/User.swift
```diff
@@ -222,12 +224,14 @@ open class User: NSObject, Credential, Mappable {
...
     open class func resetPassword(username: String, client: Client = Kinvey.sharedClient, completionHandler: VoidHandler? = nil) -> Request {
         return resetPassword(usernameOrEmail: username, client: client, completionHandler:  completionHandler)
     }
     
     /// Sends an email to the user with a link to reset the password using the `email` property.
     @discardableResult
+    @available(*, deprecated: 3.3.3, message: "Use resetPassword(usernameOrEmail:) instead")
```
heyzooi: 3.3.3 is the latest release. the next version would be 3.3.4

PR #154, opened 2016-12-06
MLIBZ-1432: mock unit tests

heyzooi:


Kinvey/Kinvey/Client.swift
```diff
@@ -72,7 +72,15 @@ open class Client: NSObject, NSCoding, Credential {
         }
     }
     
-    internal var urlSession = URLSession(configuration: URLSessionConfiguration.default) {
+    internal static let urlSessionConfiguration: URLSessionConfiguration = {
```
tejasranade: @heyzooi any particular reason to not just use `.default` here?

Kinvey/Kinvey/FileStore.swift
```diff
@@ -219,7 +219,13 @@ open class FileStore {
                 }
                 
                 let handle: (Data?, URLResponse?, Swift.Error?) -> Void = { data, response, error in
-                    if let response = response as? HTTPURLResponse , 200 <= response.statusCode && response.statusCode < 300 {
+                    if self.client.logNetworkEnabled, let response = response as? HTTPURLResponse {
```
tejasranade: @heyzooi Shouldn't this `logNetworkEnabled` check be a level lower i.e. at the request level?

Kinvey/KinveyApp/MICAuthorizationGrantViewController.swift
```diff
@@ -19,10 +19,14 @@ class MICAuthorizationGrantViewController: UIViewController {
         super.viewDidLoad()
         // Do any additional setup after loading the view, typically from a nib.
         
-        Kinvey.sharedClient.initialize(
-            appKey: "kid_rJVLE1Z5",
-            appSecret: "cd385840cbd94e2caaa8f824c2ff7f46"
-        )
+        if let appKey = ProcessInfo.processInfo.environment["KINVEY_MIC_APP_KEY"],
```
tejasranade: This is great! One suggestion - can you add instructions to the README for someone who is testing this locally vs on Travis?

heyzooi: sure! Doing it now!

Kinvey/Kinvey/Client.swift
heyzooi: I just reverted this, I think I changed to test something and forgot to revert. Thanks for notice this.

Kinvey/Kinvey/FileStore.swift
heyzooi: You're right, do you prefer to do this change in this PR or create a new ticket to change this in a separate ticket?

tejasranade: It's up to you. If there's significant work, let's do it separately. I think this PR can be merged either way.

PR #155, opened 2016-12-07
MLIBZ-1508: adding a check with a helpful error message when the user…

heyzooi:
… forgets to call super.properyMapping()


heyzooi:
@tejasranade I tried, but I couldn't figure a way to assert the crash and keep the others unit tests running

heyzooi:
maybe this could help... https://www.cocoawithlove.com/blog/2016/02/02/partial-functions-part-two-catching-precondition-failures.html
I'll try it!

PR #156, opened 2016-12-09
MLIBZ-1169: adding XCGLogger for logging

heyzooi:


Kinvey/Kinvey/DataStore.swift
```diff
@@ -214,6 +227,7 @@ open class DataStore<T: Persistable> where T: NSObject {
     @discardableResult
     open func remove(_ persistable: T, writePolicy: WritePolicy? = nil, completionHandler: IntCompletionHandler?) throws -> Request {
         guard let id = persistable.entityId else {
+            log.severe("Object Id is missing")
```
vinaygahlawat: @heyzooi I noticed here that even though we log a severe event, we do not call `fatalError(message)`, but instead throw an error.  Everywhere else where we call `log.severe(message)`, we call `fatalError(message)`.  Is there a reason we do not here?  And if so, should we just use `log.error` instead?

heyzooi: Right, since the user can catch the exception, it should be `log.error`. just changed!

PR #157, opened 2016-12-15
MLIBZ-1506: user object nil

heyzooi:


Kinvey/Kinvey/User.swift
```diff
@@ -520,7 +517,7 @@ public class User: NSObject, Credential, Mappable {
 
 }
 
-private struct UserError {
+private class UserError {
```
tejasranade: What prompted this change? Shouldn't `UserError` be a standard error object like the rest?

PR #158, opened 2016-12-20
MLIBZ-1567: custom endpoints be able to return a json array

heyzooi:


Kinvey/Kinvey/CustomEndpoint.swift
```diff
@@ -36,13 +42,28 @@ open class CustomEndpoint: NSObject {
         return request
     }
     
+    /// Executes a custom endpoint by name and passing the expected parameters.
+    @discardableResult
+    open static func executeArray(_ name: String, params: JsonDictionary? = nil, client: Client = sharedClient, completionHandler: ArrayCompletionHandler? = nil) -> Request {
```
tejasranade: @heyzooi who determines whether the request is executed as `execute()` or `executeArray()`? From your changes, looks like its the developer's responsibility to call the correct API. Instead, can we overload a single `execute()` method that uses the type of the `completionHandler` to determine what to execute? 

Can

heyzooi:
make sense. I changed to return a generic type. Do you think that we should also provide methods for the input could also be a generic type? Or maybe also a json array?

Kinvey/Kinvey/CustomEndpoint.swift
```diff
@@ -36,13 +46,77 @@ open class CustomEndpoint: NSObject {
...
+            }
+        }
+        return request
+    }
+    
+    //MARK: StaticMappable
+    
+    /// Executes a custom endpoint by name and passing the expected parameters.
+    @discardableResult
+    internal static func execute<T: StaticMappable>(_ name: String, params: JsonDictionary? = nil, client: Client = sharedClient, completionHandler: CompletionTHandler<T>? = nil) -> Request {
```
tejasranade: @heyzooi What's the difference between `Mappable` and `StaticMappable` here? That seems to be the only overload in these methods. Why is it necessary?

Kinvey/Kinvey/CustomEndpoint.swift
```diff
@@ -36,13 +46,77 @@ open class CustomEndpoint: NSObject {
         return request
     }
     
+    //MARK: Mappable
+    
+    /// Executes a custom endpoint by name and passing the expected parameters.
+    @discardableResult
+    internal static func execute<T: Mappable>(_ name: String, params: JsonDictionary? = nil, client: Client = sharedClient, completionHandler: CompletionTHandler<T>? = nil) -> Request {
```
tejasranade: These methods definitely look friendlier. I'm curious how the developer can set other request specifics such as headers or request method or `POST` body. What fields are available to custom endpoints?

Kinvey/Kinvey/CustomEndpoint.swift
heyzooi: right, just did some changes to keep a overloaded versions of `execute()`

Kinvey/Kinvey/CustomEndpoint.swift
heyzooi: right!

heyzooi: do you think we should expose headers too? request method has to be POST always and the body has to be a json object too

Kinvey/Kinvey/CustomEndpoint.swift
heyzooi: ObjectMapper documentation warns to don't extend from `BaseMappable`, so I basically would like to avoid users making this mistake forcing then to always extend from `Mappable` or `StaticMappable`

PR #159, opened 2017-01-04
When detaching objects from Realm, perform a deep copy

tejasranade:
This fixes the issue (MLIBZ-1590), of Realm being accessed from multiple threads. The root cause of the issue was the `detach` method in `RealmCache.swift`. When detaching objects, nested objects are not properly detached. As a result, any subsequent attempt to access the nested objects results in a crash.

Tests Performed:
I modified unit tests locally to test the fix. The proper test for this fix will look more like an integration test. I plan to push it out soon, but didn't want to hold up the review.

### DIFF
--- a/Kinvey/Kinvey/RealmCache.swift
+++ b/Kinvey/Kinvey/RealmCache.swift
@@ -78,18 +78,14 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
         var json:Dictionary<String, Any>
         let obj = type(of:entity).init()
         
-//        if let props = props {
         json = entity.dictionaryWithValues(forKeys: props)
-//        } else {
-//            json = entity.dictionaryWithValues(forKeys: propertyNames)
-//        }
         
         for property in json.keys {
             let value = json[property]
                 
             if let value = value as? Object {
                 
-                let nestedClassName = NSStringFromClass(type(of:value).self).components(separatedBy: "_").last!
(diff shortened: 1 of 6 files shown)


Kinvey/Kinvey/RealmCache.swift
```diff
@@ -66,16 +68,41 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
...
     }
     
-    fileprivate func newInstance<P: Persistable>(_ type: P.Type) -> P {
+    fileprivate func newInstance<P:Persistable>(_ type: P.Type) -> P {
         return type.init()
     }
-    
-    override func detach(_ entity: T) -> T {
+
+    fileprivate func detach(_ entity: Object, props: [String]) -> Object {
```
vinaygahlawat: @tejasranade Is there a reason that `detach(entity, props)` is `fileprivate` while the `detach(entity, query)` is `public`?  Also, because `detach(entity, props)` is now `fileprivate`, does it have to be deprecated?

tejasranade: Good point. `detach(entity, query)` is used in other parts of the SDK. I'd like to make it private as well, since `detach` shouldn't be an API on the `Cache` anyway. However, it'll require a better understanding of the references.

As for deprecation, I don't think it's necessary, because the `Cache` API itself is not public.

PR #166, opened 2017-01-10
MLIBZ-1575: crashing on FileStore.getInstance()

heyzooi:


Kinvey/KinveyTests/FileTestCase.swift
```diff
@@ -1077,4 +1077,10 @@ class FileTestCase: StoreTestCase {
         }
     }
     
+    func testGetInstance() {
+        let appKey = "file-get_instance-\(UUID().uuidString)"
+        let client = Client(appKey: appKey, appSecret: "unit-test")
+        let _ = FileStore.getInstance(client)
```
tejasranade: @heyzooi - Is this test complete? What are you testing for?

Kinvey/Kinvey/CacheManager.swift
```diff
@@ -81,10 +81,13 @@ internal class CacheManager: NSObject {
                     if let encryptionKey = encryptionKey {
                         realmConfiguration.encryptionKey = encryptionKey
                     }
-                    if let realm = try? Realm(configuration: realmConfiguration) , !realm.isEmpty {
+                    if let realm = try? Realm(configuration: realmConfiguration), !realm.isEmpty {
                         try! realm.write {
                             realm.deleteAll()
                         }
+                        realm.invalidate()
```
tejasranade: Can you explain the purpose of these cleanup statements?

heyzooi: invalidate() will detach any existing live objects from this Realm instance
refresh() will reevaluate the Realm instance in case another Realm instance (maybe from another Thread) made any change in the database
and the last one is just to make sure that after all this the database is indeed empty
I added all these cleanups when I was trying to clean the state of the database in the end of each unit test (tearDown) and I was seeing some weird states. Not sure we if really need this. I think I will remove that.

Kinvey/KinveyTests/FileTestCase.swift
heyzooi: basically the crash was happening because the folder for each appKey wasn't not there yet if you call `FileStore.getInstance()` so I'm creating a random appKey to check if the call would not crash

tejasranade: One suggestion is to check what you expect once the instance is created. E.g. if you can check that there's actually a Realm on disk with the correct name / prefix.

heyzooi: right, make sense, doing it now

PR #167, opened 2017-01-16
MLIBZ-1587: adding support for mapping array/list properties

heyzooi:
#### Description

This is a solution to solve the limitation around mapping arrays in your entity class.

#### Changes

- Creating new types that are literal convertible like `String`, `Int`, `Float`, `Double` and `Bool`
- Adding new `Transform` types to convert between `List` and `Array` values
- Overloading the `<-` operator in order to use the `Transform` types above

#### Tests

- Tested converting queries when running against the local cache. Queries like:
`authorNames contains 'Victor'`
`subquery(authorNames, $authorNames, $authorNames like[c] 'Vic*').$count > 0`

### DIFF
--- a/Kinvey/Kinvey/Entity.swift
+++ b/Kinvey/Kinvey/Entity.swift
@@ -11,6 +11,8 @@ import Realm
 import RealmSwift
 import ObjectMapper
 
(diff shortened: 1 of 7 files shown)


heyzooi:
@tejasranade is this one ready to merge?

PR #169, opened 2017-01-18
Restore social extras

sharplet:
#### Description

I'm currently porting an app from version 1 of the API (Objective-C) to version 3.

In commit bd0ec52d99d2e992f30ac3d258d9952e519368ac, `KCSUser+SocialExtras` was deleted. This includes the `+getAccessDictionaryFromTwitter*:` methods that my app relies on. This change adds the `KinveyKitExtras` folder to the Xcode project and podspec, and restores the social extras. At this stage the LinkedIn helper isn't included, because it relies on `KCSWebView`, which doesn't currently compile.

#### Changes

See commit log.

#### Tests

I haven't added any new tests, but my app does compile and run against this code.


heyzooi:
Thanks for your PR @sharplet!
I see this is somehow related to this PR https://github.com/Kinvey/ios-library/pull/170
We will discuss internally how we are going to approach your request and get back to you soon.

heyzooi:
@sharplet in our 1.x SDK, we handled the interaction with Twitter inside the SDK. However, we think it is best to let the app decide how to interact with Twitter to obtain the user's Twitter tokens. These can then be passed to Kinvey to [log the user in](http://devcenter.kinvey.com/ios/guides/users#usingsocialidentities). This gives you control over the Twitter interaction.

I converted the Twitter login code from our 1.x repo to Swift and it's available as a [reference implementation here](https://github.com/heyzooi/TwitterLoginWithKinvey). Let me know if it you have any trouble integrating this into your app.

sharplet:
After some further investigation, it actually looks like the code in my app that was using this was unused.

PR #170, opened 2017-01-18
Pass options dictionary through to KCSClient.initializeKinveyService()

sharplet:
#### Description
The new `Client` API doesn't provide a way to pass an options dictionary when initialising the legacy Kinvey service. This change adds a new options parameter and passes it through to `KSCClient.initializeKinveyService()`.

#### Changes
In addition to passing through the `options` dictionary, this change also introduces generic type parameters to the dictionary. This will cause a warning when using something other than an `NSString *` key, and causes the dictionary to bridge to Swift 3 as `[String: Any]`.

#### Tests
No new tests were added.

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -203,7 +203,7 @@ open class Client: NSObject, NSCoding, Credential {
     }
     
     /// Initialize a `Client` instance with all the needed parameters.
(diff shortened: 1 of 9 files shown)


heyzooi:
Thanks for your contribution @sharplet!
We have planned already in our road map to remove the legacy code from v1, which means remove the Obj-C classes that we still need to use for a few things a port them all to Swift. After we conclude that you don't need to use the options dictionary anymore since we are creating specific properties for each configuration.
This should coming soon. So stay tuned.

I'm curious, what are the options that you are using, so we can make sure that we will cover that.

sharplet:
Thanks for the quick response. The keys we're using are `KCS_TWITTER_CLIENT_KEY` & `KCS_TWITTER_CLIENT_SECRET`, and `KCS_FACEBOOK_APP_KEY`.

tejasranade:
Closing since this is resolved by other PRs.

PR #171, opened 2017-01-23
Deserialize active user as the correct custom user type

sharplet:
When using the asynchronous login and logout APIs, the active user is correctly mapped using the configured custom `User` subclass. However on initialisation, when the active user is deserialised from user defaults, the custom subclass is ignored. This can result in surprising bugs or crashes when callers expect their custom `User` subclassed to be used on app launch.

Because `Client.userType` is specified dynamically, rather than statically, it's not possible to parameterise the `Mapper` using the dynamic type (dynamic types can't be used in static declarations). Thus in order to configure a mapper of dynamic type, we need to use type erasure: this is achieved using the `AnyMapper` type.

Swift properties can't have generic type parameters in the same way that functions can, so in order to parameterise the setter a new generic function `setUserType(_:)` has been introduced. This internally constructs the type-erased `userMapper`, which can later be accessed when deserialising from user defaults, and the mapped object dynamically cast back to the expected `User` subclass.

I don't believe there's a way to make this fix a non-breaking change.

### DIFF
--- a/Kinvey/Kinvey.xcodeproj/project.pbxproj
+++ b/Kinvey/Kinvey.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		4A68248E1E3696DA0018B4F9 /* AnyMapperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A68248D1E3696DA0018B4F9 /* AnyMapperTests.swift */; };
+		4A6824901E3696E20018B4F9 /* AnyMapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A68248F1E3696E20018B4F9 /* AnyMapper.swift */; };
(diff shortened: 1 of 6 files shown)


heyzooi:
@sharplet thanks once again for your contributions!
We really liked your solution of erasing types and we will probably use for a few things in the future!

We were working in parallel to remove all the legacy code and we fixed this problem in this PR https://github.com/Kinvey/ios-library/pull/172 (already merged into develop branch)

Here's the line of code that should fix your problem
https://github.com/Kinvey/ios-library/blob/3bc8306e9632b513a23bec958ef8e422af975926/Kinvey/Kinvey/Client.swift#L209

Please let us know if this fixed your problem.

sharplet:
I did notice that today actually. I tried `develop` but I think there might have been a podspec issue, as I couldn't get it to compile at the time. I'll go ahead and close this though, I wasn't aware of `User.init(JSON:)`, looks good!

Issue #173, opened 2017-01-26
Kinvey.Error doesn't conform to LocalizedError


sharplet:
### Description

`Kinvey.Error.localizedDescription` gives an unhelpful error message:

![simulator screen shot jan 26 2017 10 13 07 am](https://cloud.githubusercontent.com/assets/147686/22336908/b91dbdde-e3b1-11e6-877c-d8a3f9e1b3c7.png)

This issue would be fixed by adding `LocalizedError` conformance. See https://github.com/apple/swift-evolution/blob/master/proposals/0112-nserror-bridging.md#proposed-solution.

### Steps to reproduce 

Trigger a `Kinvey.Error`, for example by logging in with the wrong password.

### Kinvey SDK Version

Using Kinvey from `develop`.

### Tool Versions (XCode / Swift / iOS / Device)

n/a

heyzooi:
@sharplet thanks again for your contributions! Great suggestion!
We just pushed a new PR for that https://github.com/Kinvey/ios-library/pull/174
Should be merged soon and will probably be in the next release.

PR #174, opened 2017-01-26
MLIBZ-1394: Better error message

heyzooi:
#### Description

Better error messages for `Kinvey.Error` objects as described in the issue https://github.com/Kinvey/ios-library/issues/173

#### Changes

`Kinvey.Error` now conforms to `LocalizedError`

#### Tests

Unit Test changed accordingly

### DIFF
--- a/Kinvey/Kinvey/Error.swift
+++ b/Kinvey/Kinvey/Error.swift
@@ -9,7 +9,7 @@
 import Foundation
 
 /// Enum that contains all error types in the library.
-public enum Error: Swift.Error, CustomStringConvertible, CustomDebugStringConvertible {
+public enum Error: Swift.Error, LocalizedError, CustomStringConvertible, CustomDebugStringConvertible {
     
     /// Constant for 401 responses where the credentials are not enough to complete the request.
     public static let InsufficientCredentials = "InsufficientCredentials"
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/Error.swift
```diff
@@ -89,6 +89,14 @@ public enum Error: Swift.Error, CustomStringConvertible, CustomDebugStringConver
         }
     }
     
+    public var errorDescription: String? {
+        return description
+    }
+    
+    public var failureReason: String? {
+        return description
```
sharplet: Is there much value having `errorDescription` and `failureReason` be the same thing?

heyzooi: not really, but if someone's code is relying in one of them it could be useful. Also, strictly speaking, `errorDescription` and `failureReason` have similar meaning. One thing that we can do in the future is use the `errorDescription` as overview of the error and leave the `failureReason` to describe in more details the reason why error happened. Any thoughts? @tejasranade what's your opinion?

sharplet: Hmm, perhaps I was mistaken: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html#//apple_ref/doc/uid/TP40001806-CH202-SW4

> A brief sentence that explains the reason why the error occurred. It is typically part of the error description.

Also:

> An example can help to clarify the relationship between error description and failure reason. An error object has an error description of “File could not be saved because the disk is full.” The accompanying failure reason is “The disk is full.”

heyzooi: right! so in this case we should actually invert the order... `errorDescription` should have a bigger explanation and `failureReason` a short version containing only the reason

PR #176, opened 2017-02-02
MLIBZ-1167: Geolocation Queries

heyzooi:
#### Description

Solves the problem of perform geolocation queries in the backend and locally

#### Changes

- NSPredicate -> MongoDB adaptor
- Local cache

#### Tests

- Test geolocation queries with circles with a radio
- Test geolocation queries with polygons (which includes the use case of boxes / rectangles)

### DIFF
--- a/Kinvey/Kinvey/Geolocation.swift
+++ b/Kinvey/Kinvey/Geolocation.swift
@@ -1,0 +1,106 @@
+//
+//  Geolocation.swift
+//  Kinvey
+//
+//  Created by Victor Hugo on 2017-02-01.
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+import ObjectMapper
+import CoreLocation
+import MapKit
+
+open class GeoPoint: Object, Mappable {
+    
+    open dynamic var latitude: CLLocationDegrees = 0.0
+    open dynamic var longitude: CLLocationDegrees = 0.0
+    
+    public convenience required init?(map: Map) {
+        guard let _: Double = map["latitude"].value(), let _: Double = map["longitude"].value() else {
+            return nil
+        }
+        self.init()
+    }
+    
+    public convenience init(latitude: CLLocationDegrees, longitude: CLLocationDegrees) {
+        self.init()
+        self.latitude = latitude
+        self.longitude = longitude
+    }
+    
+    public convenience init(coordinate: CLLocationCoordinate2D) {
+        self.init(latitude: coordinate.latitude, longitude: coordinate.longitude)
+    }
+    
+    public convenience init(location: CLLocation) {
+        self.init(latitude: location.coordinate.latitude, longitude: location.coordinate.longitude)
+    }
+    
+    public convenience init(_ array: [CLLocationDegrees]) {
+        self.init(latitude: array[1], longitude: array[0])
+    }
+    
+    public func mapping(map: Map) {
+        latitude <- map["latitude"]
+        longitude <- map["longitude"]
+    }
+    
+}
+
+class GeoPointTransform: TransformOf<GeoPoint, [CLLocationDegrees]> {
+    
+    init() {
+        super.init(fromJSON: { (array) -> GeoPoint? in
+            if let array = array, array.count == 2 {
+                return GeoPoint(array)
+            }
+            return nil
+        }, toJSON: { (geopoint) -> [CLLocationDegrees]? in
+            if let geopoint = geopoint {
+                return [geopoint.longitude, geopoint.latitude]
+            }
+            return nil
+        })
+    }
+    
+}
+
+/// Override operator used during the `propertyMapping(_:)` method.
+public func <- (left: inout GeoPoint, right: (String, Map)) {
+    kinveyMappingType(left: right.0, right: right.1.currentKey!)
+    left <- (right.1, GeoPointTransform())
+}
+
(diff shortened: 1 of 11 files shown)


Kinvey/Kinvey/Geolocation.swift
```diff
@@ -0,0 +1,112 @@
...
+    dynamic var longitude: CLLocationDegrees = 0.0
+    
+    public static func objectForMapping(map: Map) -> BaseMappable? {
+        guard let _: Double = map["latitude"].value(), let _: Double = map["longitude"].value() else {
+            return nil
+        }
+        return GeoPoint()
+    }
+    
+    public convenience init(latitude: Double, longitude: Double) {
```
tejasranade: @heyzooi if we have the constructor for `CLLocation`, what is the reason to have this one separately? I prefer we keep one way of doing things when possible.

Kinvey/Kinvey/Geolocation.swift
```diff
@@ -0,0 +1,112 @@
...
+        self.longitude = coordinate.longitude
+    }
+    
+    public convenience init(location: CLLocation) {
+        self.init()
+        self.latitude = location.coordinate.latitude
+        self.longitude = location.coordinate.longitude
+    }
+    
+    public convenience init(_ array: [Double]) {
```
tejasranade: Same comment as above. Exposing this constructor adds unnecessary work e.g. we have to document which value in the array is lat and which is long, etc.

Kinvey/Kinvey/Persistable.swift
```diff
@@ -55,19 +56,19 @@ public func <- <T>(left: inout T!, right: (String, Map)) {
 }
 
 /// Override operator used during the `propertyMapping(_:)` method.
-public func <- <T: Mappable>(left: inout T, right: (String, Map)) {
+public func <- <T: BaseMappable>(left: inout T, right: (String, Map)) {
```
tejasranade: @heyzooi what is the reason for this change?

Makefile
```diff
@@ -24,7 +24,7 @@ test: test-ios
 
 	
 test-ios:
-	xcodebuild -workspace Kinvey.xcworkspace -scheme Kinvey -destination 'platform=iOS Simulator,name=iPhone 7,OS=10.2' -enableCodeCoverage YES test
+	xcodebuild -workspace Kinvey.xcworkspace -scheme Kinvey -destination 'platform=iOS Simulator,name=iPhone 7,OS=10.1' -enableCodeCoverage YES test
```
tejasranade: @heyzooi why was this moved down?

Kinvey/Kinvey/RealmCache.swift
```diff
@@ -123,7 +154,55 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
...
     func detach(_ results: RealmSwift.Results<Entity>, query: Query?) -> [T] {
-        return detach(results.map { $0 as! T }, query: query)
+        var results: [T] = results.map { $0 as! T }
+        if let predicate = query?.predicate as? NSComparisonPredicate,
+            let keyPathConstantValuePair = keyPathConstantValuePair(predicate: predicate),
+            let constantValue = keyPathConstantValuePair.constantValueExpression.constantValue,
+            constantValue is MKCircle || constantValue is MKPolygon
+        {
+            if let circle = constantValue as? MKCircle {
+                let center = CLLocation(latitude: circle.coordinate.latitude, longitude: circle.coordinate.longitude)
```
tejasranade: I don't full comprehend this change. Let's discuss. On first glance, looks odd in the `RealmCache` class.

Kinvey/Kinvey/Geolocation.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+import ObjectMapper
+import CoreLocation
+import MapKit
+
+open class GeoPoint: Object, StaticMappable {
```
heyzooi: for this one, not really... agree, changing now

Kinvey/Kinvey/Geolocation.swift
heyzooi: forgot to summarize the constructors, doing it now... that's why I created this kinda agnostic constructor

Kinvey/Kinvey/Geolocation.swift
heyzooi: same reason... please check the next commit

Kinvey/Kinvey/Persistable.swift
heyzooi: to also cover `StaticMappable` which also extends from `BaseMappable`

Makefile
heyzooi: this is to fix a problem in Travis CI... don't know why, but Travis CI has 2 iOS Simulator with 10.2 and for that reason the script ask you to specify the UUID of the Simulator (which can also change from time to time)

Kinvey/Kinvey/RealmCache.swift
heyzooi: sure... but basically Realm does not support geolocation queries, so we need to do it manually after fetch the results

Makefile
heyzooi: here's an example: https://travis-ci.org/Kinvey/ios-library/builds/198941783

PR #177, opened 2017-02-03
MLIBZ-1167: MongoDB adaptor conversion to Swift

heyzooi:
#### Description

Converting the MongoDB adaptor to Swift

#### Changes

- MongoDB adaptor

#### Tests

- Same unit tests already exists


heyzooi:
@tejasranade 

* The code coverage is complaining probably because Travis failed. I'll double check that.

* Agree, we should keep this as a separate module. My plan is to submit this change as a PR to the original author

* Travis is now fixed

PR #178, opened 2017-02-07
MLIBZ-1671: Result enum

heyzooi:
#### Description

This is a proposal to discuss.
Result enum solves the problem of having a never reached code, for example:

```
CustomEndpoint.execute("echo", params: CustomEndpoint.Params(params)) { (response: [EchoType]?, error: Swift.Error?) in
    if let response = response {
        // reachable
    } else if let error = error {
        // reachable
    } else {
        // will never be reached!
    }
}
```

and now could  be done this way:

```
CustomEndpoint.execute("echo", params: CustomEndpoint.Params(params)) { (result: Result<[EchoType]>) in
    switch result {
    case .success(let response):
        // handle success
    case .failure(let error):
        // handle failure
    }
}
```

#### Changes

For now, until we agree that we should do this, it's implemented only for `CustomEndpoint`

#### Tests

CustomEndpoint's unit tests changed accordingly


thomasconner:
How is this handled in other Swift libraries that have callbacks?

heyzooi:
I believe this concept was introduced by this library https://github.com/antitypical/Result
Apple's code is using the same approach that we are using (optional params result and error where only one should be not null), but the Swift community is finding this new approach better because makes the code more readable and fail safe since it forces the user to handle the `.success` case and the `.error` case.

Here's some articles showing how people are using this new approach:

https://nomothetis.svbtle.com/error-handling-in-swift
https://www.cocoawithlove.com/blog/2016/08/21/result-types-part-one.html
https://www.natashatherobot.com/swift-generics-box/
http://tech.domain.com.au/2016/03/result-oriented-swift/
https://christiantietze.de/posts/2015/05/nicest-result-enum-ever/

heyzooi:
rejected in favor of https://github.com/Kinvey/ios-library/pull/199

PR #179, opened 2017-02-09
MLIBZ-1667: WIP fixed file json mapping

tejasranade:
#### Description
Fixes MLIBZ-1667 - the ACL properties set on a file object are not being POSTed to the backend.

#### Changes
* The `File` class now extends `Entity`. This allows us to define JSON mappings for it like other kinvey entities.
* All serialization and deserialization of `File` object is via JSON mapper.
* (Proposed) - since `Entity` maps `_id` already, is there any value to having a separate `fileId`? The proposal is to remove `fileId`, although that'll be a breaking change.
* (Proposed) - current file APIs pass the `File` object as a reference. In all of our other data APIs, we build new instances in response to `save` or `fetch` calls. Is there a reason to use `File` as references? The proposal is to treat `File` like any data object. The instance received in the completion block is a new object.

#### Tests
* Current unit tests pass with minor changes.

### DIFF
--- a/Kinvey/Kinvey/FileStore.swift
+++ b/Kinvey/Kinvey/FileStore.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import PromiseKit
+import ObjectMapper
 
 #if os(iOS)
     import UIKit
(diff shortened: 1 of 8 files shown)


Kinvey/Kinvey/FileStore.swift
```diff
@@ -148,16 +121,18 @@ open class FileStore {
...
+                        
+                        fulfill((file: newFile, skip: nil))
                     } else {
                         reject(buildError(data, response, error, self.client))
                     }
                 }
             }
             
-            if let _ = file.fileId {
+            if let _ = file.entityId {
```
tejasranade: @heyzooi these all need to go back to `fileId`.

PR #181, opened 2017-02-16
MLIBZ-1521: default transform for arrays

heyzooi:
#### Description

Convenient overload for <- operator to handle List<T> types

#### Changes

* Overload the <- operator to handle list of mappable objects

#### Tests

Current unit tests should also cover that

### DIFF
--- a/Kinvey/Kinvey/Persistable.swift
+++ b/Kinvey/Kinvey/Persistable.swift
@@ -10,8 +10,12 @@ import Foundation
 import CoreData
 import ObjectMapper
 import CoreLocation
+import RealmSwift
 
 public typealias Map = ObjectMapper.Map
+public typealias Mappable = ObjectMapper.Mappable
+public typealias StaticMappable = ObjectMapper.StaticMappable
+
 infix operator <- : DefaultPrecedence
 
 /// Protocol that turns a NSObject into a persistable class to be used in a `DataStore`.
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/Persistable.swift
```diff
@@ -93,6 +94,40 @@ public func <- <Transform: TransformType>(left: inout Transform.Object!, right:
     kinveyMappingType(left: right.0, right: right.1.currentKey!)
     left <- (right.1, right.2)
 }
+  
+class ListValueTransform<T: RealmSwift.Object>: TransformOf<List<T>, [JsonDictionary]> where T: BaseMappable {
```
tejasranade: I think we should prefer using `TransformOf` to do quick transformations as described [here](https://github.com/Hearst-DD/ObjectMapper#transformof). 
If we need to extend (such as what you're doing here), `TransformType` seems more appropriate.

heyzooi: `TransformOf` actually implements `TransformType` plus it checks the type inside `fromJSON()`, so no big difference despite the fact that using `TransformOf` you end up writing a little less code.
Let me know if you really thinks that using `TransformType` would be better.

tejasranade: That's fine. For public docs, it'll be nice to prescribe one way of writing transforms.

PR #183, opened 2017-02-23
Realm and RealmSwift (Kinvey Dependencies) Update breaks Collection.Find() - Work Around

Sagar-AppLogics:
Realm and RealmSwift have both been updated from 2.4.2, to 2.4.3 few weeks AFTER this Kinvey iOS SDK version 3.3.8 was released. Now when a new project has pods installed or a existing project updates its Pods, the version of Realm and RealmSwift has updated, after which point. Calls to Collection.find() fails. This is due to the update of Realm/RealmSwift, when explicitly stating 2.4.2 as the Realm and RealmSwift version with the latest (3.3.8) SDK, explicitly stating the versions for both these dependencies is the only way to make it work,.
I will call this a workaround although I don'y know if specifying a version for your dependency comes under the category of workaround or normal practice. I don't know exactly if the fault lies with Kinvey or Relam/RealmSwift but this pull request will probably be required until the issue is fixed at either end. If its a Kinvey issue it may buy the Kinvey Engineers some more time to figure out if the problem. or keep the Kinvey SDK's dependency versions for the both Realm and RealmSwift fixed until the time a version is released that functions correctly release?

Note: this will prevent users from being unable to access their **_MAIN_** data on the iOS platform
Note: I have only changed dependencies for CocoaPod's podspec file, any other dependancy managed configurations will probably cause it to fail so someone may have to tweak for that.

### DIFF
--- a/Kinvey.podspec
+++ b/Kinvey.podspec
@@ -141,8 +141,8 @@ Pod::Spec.new do |s|
 	
   s.dependency "PromiseKit", "~> 4.0"
   s.dependency "KeychainAccess", "~> 3.0"
-  s.dependency "Realm", "~> 2.0"
-  s.dependency "RealmSwift", "~> 2.0"
+  s.dependency "Realm", "= 2.4.2"
+  s.dependency "RealmSwift", "= 2.4.2"
   s.dependency "ObjectMapper", "~> 2.0"
   s.dependency "XCGLogger", "~> 4.0"
 


tejasranade:
@Sagar-AppLogics - thanks for the PR. The change makes sense. Can you make the following adjustments - 

- Before we accept your first contribution, we require that you sign our [CLA](https://goo.gl/forms/spZb2rXhC6I6zOxw1). Please fill out and submit the CLA so we can merge your PR.

- Instead of pinning the Realm version to 2.4.2, it's better that we set it to 2.4.2 or higher i.e.
`"~ 2.4.2"` instead of `"= 2.4.2"`

Thanks!

heyzooi:
closing since was solved by the latest release

PR #186, opened 2017-03-02
MLIBZ-1700: performance tests and minor improvements

heyzooi:
#### Description

Testing a pre-defined set of data to test performance fetching data

#### Changes

* Minor changes in the code to improve performance

#### Tests

* Unit tests to be able to collect performance numbers

### DIFF
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -408,9 +408,8 @@ open class DataStore<T: Persistable> where T: NSObject {
     }
 
     /// Clear all data for the collection attached to the DataStore.
-    open func clearCache() {
-        cache?.removeAllEntities()
-        sync?.removeAllPendingOperations()
+    open func clearCache(query: Query? = nil) {
+        cache?.clear(query: query)
     }
 
 }
--- a/Kinvey/Kinvey/FileStore.swift
+++ b/Kinvey/Kinvey/FileStore.swift
@@ -535,23 +535,17 @@ open class FileStore {
     open func find(_ query: Query = Query(), ttl: TTL? = nil, completionHandler: FileArrayCompletionHandler? = nil) -> Request {
         let request = client.networkRequestFactory.buildBlobQueryFile(query, ttl: ttl)
         Promise<[File]> { fulfill, reject in
-            request.execute({ (data, response, error) -> Void in
-                if let response = response , response.isOK,
-                    let jsonArray = self.client.responseParser.parseArray(data)
+            request.execute { (data, response, error) -> Void in
+                if let response = response,
+                    response.isOK,
+                    let jsonArray = self.client.responseParser.parseArray(data),
+                    let files = [File](JSONArray: jsonArray)
                 {
-                    var files: [File] = []
-                    for json in jsonArray {
-                        //let file = File()
-                        //self.fillFile(file, json: json)
-                        if let file = Mapper<File>().map(JSON: json) {
-                            files.append(file)
-                        }
-                    }
                     fulfill(files)
                 } else {
                     reject(buildError(data, response, error, self.client))
                 }
-            })
+            }
         }.then { files in
             completionHandler?(files, nil)
         }.catch { error in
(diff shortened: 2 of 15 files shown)


Kinvey/Kinvey/DataStore.swift
```diff
@@ -409,8 +409,7 @@ open class DataStore<T: Persistable> where T: NSObject {
 
     /// Clear all data for the collection attached to the DataStore.
     open func clearCache() {
```
tejasranade: @heyzooi  Not directly related to this change, but shouldn't `clearCache()` optionally take a query argument?

Kinvey/Kinvey/FileStore.swift
```diff
@@ -509,23 +509,17 @@ open class FileStore {
...
         let request = client.networkRequestFactory.buildBlobQueryFile(query, ttl: ttl)
         Promise<[File]> { fulfill, reject in
-            request.execute({ (data, response, error) -> Void in
-                if let response = response , response.isOK,
-                    let jsonArray = self.client.responseParser.parseArray(data)
+            request.execute { (data, response, error) -> Void in
+                if let response = response,
+                    response.isOK,
+                    let jsonArray = self.client.responseParser.parseArray(data),
+                    let files = [File](JSONArray: jsonArray)
```
tejasranade: @heyzooi what mapping function is getting used here?

Kinvey/Kinvey/RealmCache.swift
```diff
@@ -449,6 +449,14 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
         }
     }
     
+    override func clear() {
```
tejasranade: @heyzooi similar to my previous comment, there should be an optional query parameter here.

Kinvey/KinveyTests/PerformanceProductTestCase.swift
```diff
@@ -0,0 +1,447 @@
...
+    
+}
+
+class PerformanceProductTestCase: KinveyTestCase {
+    
+    var productsJsonArray: [JsonDictionary]?
+    
+    lazy var fileStore = FileStore.getInstance()
+    
+    lazy var productsJsonFile: File? = {
```
tejasranade: @heyzooi side note - if you plan to commit the product json file, it might be better to test against a smaller number of entities, like 100k or so. Otherwise our repo will get bloated again, and also the test will take long to complete.

Another suggestion is to turn this test off by default, to avoid travis taking too long to finish.

Kinvey/KinveyTests/QueryTest.swift
```diff
@@ -13,7 +13,8 @@ import MapKit
 class QueryTest: XCTestCase {
     
     func encodeQuery(_ query: Query) -> String {
-        return query.queryParams.urlQueryEncoded.removingPercentEncoding!
+        var urlComponents = URLComponents(url: URL(string: "parse://")!, resolvingAgainstBaseURL: false)!
```
tejasranade: What is the significance of the `parse://` prefix?

Kinvey/Kinvey/FileStore.swift
heyzooi: the the regular `mapping()`

Kinvey/KinveyTests/QueryTest.swift
heyzooi: nothing really... it's just any url scheme so I can get the query params

Kinvey/KinveyTests/PerformanceProductTestCase.swift
heyzooi: right... I'm already doing this... the file is actually stored in a Kinvey backend (as a file) and the unit test is also disabled by default so Travis will not run.

Kinvey/Kinvey/RealmCache.swift
heyzooi: sorry, not sure what you mean?

Kinvey/Kinvey/DataStore.swift
heyzooi: that's a good idea, I will add this feature

heyzooi:
@tejasranade it should be good to go now

PR #189, opened 2017-03-15
MLIBZ-1713: throwing a proper error if the push configuration is missing in the backend

heyzooi:
#### Description

Missing APNS push configuration in the console should return an error

#### Changes

* Building the right proper error

#### Tests

* Unit test mocking an error response from the backend


Kinvey/Kinvey.xcodeproj/project.pbxproj
```diff
@@ -2764,6 +2764,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.kinvey.KinveyAppTests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_VERSION = 3.0;
```
tejasranade: @heyzooi was this intentional?

heyzooi: yes, one of our unit test targets was missing the swift version param which causes Xcode keeping ask to migrate every time you open Xcode, so this will stop.

PR #191, opened 2017-03-20
MLIBZ-1660: Local query should not crash if field not found on entity

heyzooi:
#### Description

Local query should not crash if a field is not mapped in the entity class

#### Changes

* Make the local cache ignore `Invalid property name` errors

#### Tests

* Unit Test runs a local query with a non-mapped field

### DIFF
--- a/Kinvey/Kinvey/RealmCache.swift
+++ b/Kinvey/Kinvey/RealmCache.swift
@@ -237,11 +237,21 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
         }
     }
     
-    fileprivate func results(_ query: Query) -> RealmSwift.Results<Entity> {
+    fileprivate func results(_ query: Query) -> AnyRandomAccessCollection<Entity> {
         log.verbose("Fetching by query: \(query)")
         var realmResults = self.realm.objects(self.entityType)
         if let predicate = query.predicate {
-            realmResults = realmResults.filter(translate(predicate: predicate))
+            if let exception = tryBlock({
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/RealmCache.swift
```diff
@@ -386,7 +396,7 @@ internal class RealmCache<T: Persistable>: Cache<T> where T: NSObject {
         var result = 0
         executor.executeAndWait {
             if let query = query {
-                result = self.results(query).count
+                result = Int(self.results(query).count)
```
tejasranade: @heyzooi Can `self.results(query)` ever be `nil`? If not, this is fine.

Kinvey/KinveyTests/SyncStoreTests.swift
```diff
@@ -1342,4 +1342,23 @@ class SyncStoreTests: StoreTestCase {
         
     }
     
+    func testQueryWithPropertyNotMapped() {
+        let query = Query(format: "propertyNotMapped == %@", 10)
+        
+        weak var expectationFind = expectation(description: "Find")
+        
+        store.find(query) { persons, error in
```
heyzooi: right, but since the problem was happening in the Realm layer, I first wrote a unit test that actually does reproduce the problem and then I re-run the same test again to make sure the unit test now passes without error.

Kinvey/Kinvey/RealmCache.swift
heyzooi: not, it can be empty, which will return a `count` equals to `0`, but not nil

PR #192, opened 2017-03-22
MLIBZ-1725: upload from input stream

heyzooi:
#### Description

Upload from input stream done properly

#### Changes

* Change the way we were uploading data if the data source was an input stream

#### Tests

* Unit Test to upload from an input stream

### DIFF
--- a/Kinvey/Kinvey/TaskProgressRequest.swift
+++ b/Kinvey/Kinvey/TaskProgressRequest.swift
@@ -125,7 +125,7 @@ class TaskProgressRequest: NSObject {
         {
             let progressStatus = ProgressStatus(task)
             if (progressStatus.countOfBytesSent == progressStatus.countOfBytesExpectedToSend && progressStatus.countOfBytesReceived >= 0 && progressStatus.countOfBytesExpectedToReceive != 0) ||
-                (progressStatus.countOfBytesSent >= 0 && progressStatus.countOfBytesExpectedToSend > 0)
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/TaskProgressRequest.swift
```diff
@@ -125,7 +125,7 @@ class TaskProgressRequest: NSObject {
         {
             let progressStatus = ProgressStatus(task)
             if (progressStatus.countOfBytesSent == progressStatus.countOfBytesExpectedToSend && progressStatus.countOfBytesReceived >= 0 && progressStatus.countOfBytesExpectedToReceive != 0) ||
-                (progressStatus.countOfBytesSent >= 0 && progressStatus.countOfBytesExpectedToSend > 0)
+                (progressStatus.countOfBytesSent >= 0 && (progressStatus.countOfBytesExpectedToSend > 0 || progressStatus.countOfBytesExpectedToSend == -1))
```
tejasranade: @heyzooi under what circumstances is `countOfBytesExpectedToSend` = 0 or -1 respectively? I suggest adding some comments to this code so it leaves some indication of when these cases are hit.

heyzooi: `countOfBytesExpectedToSend` is `0` when the data is actually empty and it could can also be `-1` if the upload data is actually and unknown size stream

PR #193, opened 2017-03-23
MLIBZ-1723: making push operations able to run in parallel

heyzooi:
#### Description

Making push operations able to run in parallel
Performance tests jump from around `30s` to `5s` to push 100 records

#### Changes

* Not making pending operations depending each other so they can run in parallel

#### Tests

* Unit Test to make sure that only one pending operation is stored by entity

### DIFF
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -230,7 +230,7 @@ open class DataStore<T: Persistable> where T: NSObject {
             log.error("Object Id is missing")
             throw Error.objectIdMissing
         }
-        return removeById(id, writePolicy:writePolicy, completionHandler: completionHandler)
+        return remove(byId: id, writePolicy:writePolicy, completionHandler: completionHandler)
     }
     
     /// Deletes a list of records.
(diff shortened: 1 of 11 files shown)


Kinvey/Kinvey/DataStore.swift
```diff
@@ -242,12 +242,18 @@ open class DataStore<T: Persistable> where T: NSObject {
...
     
     /// Deletes a record using the `_id` of the record.
     @discardableResult
+    @available(*, deprecated: 3.4.0, message: "Please use `remove(byId:)` instead")
     open func removeById(_ id: String, writePolicy: WritePolicy? = nil, completionHandler: IntCompletionHandler?) -> Request {
+        return remove(byId: id, writePolicy: writePolicy, completionHandler: completionHandler)
+    }
+    
+    @discardableResult
+    open func remove(byId id: String, writePolicy: WritePolicy? = nil, completionHandler: IntCompletionHandler?) -> Request {
```
tejasranade: @heyzooi this needs an API reference.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -258,7 +264,13 @@ open class DataStore<T: Persistable> where T: NSObject {
     
     /// Deletes a list of records using the `_id` of the records.
     @discardableResult
+    @available(*, deprecated: 3.4.0, message: "Please use `remove(byIds:)` instead")
     open func removeById(_ ids: [String], writePolicy: WritePolicy? = nil, completionHandler: IntCompletionHandler?) -> Request {
+        return remove(byIds: ids, writePolicy: writePolicy, completionHandler: completionHandler)
+    }
+    
+    @discardableResult
+    open func remove(byIds ids: [String], writePolicy: WritePolicy? = nil, completionHandler: IntCompletionHandler?) -> Request {
```
tejasranade: Same. Needs API reference.

Kinvey/Kinvey/RealmSync.swift
```diff
@@ -56,6 +56,12 @@ class RealmSync<T: Persistable>: SyncType where T: NSObject {
         log.verbose("Saving pending operation: \(pendingOperation)")
         executor.executeAndWait {
             try! self.realm.write {
+                if !pendingOperation.collectionName.isEmpty,
```
tejasranade: @heyzooi this makes sense. When you push, how many pending operations are retrieved at one time to push? Is the the entire queue, or is there a batch size?

heyzooi: the entire queue, but Realm does not load the entire result set in memory. it does incrementally.

Kinvey/Kinvey/DataStore.swift
heyzooi: fixing...

Kinvey/Kinvey/DataStore.swift
heyzooi: fixing...

PR #198, opened 2017-04-10
MLIBZ-1735: ping call

heyzooi:
#### Description

Adding missing ping call that we used to have in v1 of the library

#### Changes

* `ping()` method added to `Client`

#### Tests

* Unit Test to test successful and failure cases

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import ObjectMapper
+import PromiseKit
 
 private let lockEncryptionKey = NSLock()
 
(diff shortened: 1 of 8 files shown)


Kinvey/Kinvey/Client.swift
```diff
@@ -266,10 +267,62 @@ open class Client: Credential {
         return Client.fileURL(appKey: self.appKey!, tag: tag)
     }
     
-    public func encode(with aCoder: NSCoder) {
```
thomasconner: Why was this removed?

heyzooi: this is a legacy code when we had some support for Obj-C, so no need to keep it.

thomasconner: Ok

PR #201, opened 2017-04-21
MLIBZ-1788: user refresh

heyzooi:
#### Description

Missing feature from v1 which allows an user instance to refresh data

#### Changes

* Added a new method in the `User` class

#### Tests

* Unit Test added with mocking response

### DIFF
--- a/Kinvey/Kinvey/User.swift
+++ b/Kinvey/Kinvey/User.swift
@@ -360,7 +360,7 @@ open class User: NSObject, Credential, Mappable {
         return request
     }
     
-    /// Gets a `User` instance using the `userId` property.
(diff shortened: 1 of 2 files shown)


Kinvey/Kinvey/User.swift
```diff
@@ -360,6 +360,28 @@ open class User: NSObject, Credential, Mappable {
         return request
     }
     
+    /// Gets a `User` instance using the `userId` property.
```
tejasranade: @heyzooi this note doesn't seem to match the method signature.

heyzooi: right, fixing it now

PR #203, opened 2017-04-26
MLIBZ-1793: Improve Test Coverage

heyzooi:
#### Description

Improve Test Coverage to 90%

#### Changes

* Removing code not used
* Improving Unit Tests to cover more cases

#### Tests

* Some new tests added to improve coverage


Kinvey/Kinvey/Error.swift
```diff
@@ -162,8 +162,13 @@ public enum Error: Swift.Error, LocalizedError, CustomStringConvertible, CustomD
         return nil
     }
     
-    static func buildUnknownError(httpResponse: HTTPURLResponse?, data: Data?, error: String) -> Error {
-        return unknownError(httpResponse: httpResponse, data: data, error: error)
+    internal var responseBodyJsonDictionary: JsonDictionary? {
```
tejasranade: @heyzooi what prompted this change? Was the `buildUnknownError` method unused?

Kinvey/Kinvey/HttpRequest.swift
```diff
@@ -307,9 +313,7 @@ internal class HttpRequest: TaskProgressRequest, Request {
         self.endpoint = Endpoint.url(url: request.url!)
         self.client = client
         
-        if let authorization = request.value(forHTTPHeaderField: HttpHeader.authorization(credential: nil).name) {
-            self.credential = HttpHeaderCredential(authorization)
-        } else {
+        if request.value(forHTTPHeaderField: HttpHeaderKey.authorization.rawValue) == nil {
```
tejasranade: @heyzooi If the value is not `nil`, will `self.credential` be set correctly? The old and new code don't seem to match up, hence the question.

Kinvey/Kinvey/User.swift
```diff
@@ -584,8 +575,10 @@ open class User: NSObject, Credential, Mappable {
         let request = client.networkRequestFactory.buildUserSave(user: self, newPassword: newPassword)
         Promise<U> { fulfill, reject in
             request.execute() { (data, response, error) in
-                if let response = response , response.isOK, let user = client.responseParser.parseUser(data) as? U {
-                    client.activeUser = user
+                if let response = response, response.isOK, let user = client.responseParser.parseUser(data) as? U {
+                    if user.userId == client.activeUser?.userId {
```
tejasranade: @heyzooi if this condition fails (the only reason I can imagine is because the active user changed while the `save` was in progress), then shouldn't we error out?

Kinvey/Kinvey/Error.swift
heyzooi: yes, unused code

heyzooi:
@tejasranade don't merge yet, I still working on it to reach 90%

Kinvey/Kinvey/HttpRequest.swift
heyzooi: if it's not `nil` it's because you already have the `Authorization` header, which means that we don't need to set the `credential` for this request object.

Kinvey/Kinvey/User.swift
heyzooi: this is to avoid set the wrong user as an active user if you try to save an user that is not yourself

PR #210, opened 2017-05-25
MLIBZ-1796: custom file type

heyzooi:
#### Description

Allow users to extend from `File` and create their own properties for files.

#### Changes

* Changes on `FileCache`, `RealmFileCache`, `CacheStore` and `CacheManager`

#### Tests

* Minor changes in the unit tests for Files to cover the use case

### DIFF
--- a/Kinvey/Kinvey/FileStore.swift
+++ b/Kinvey/Kinvey/FileStore.swift
@@ -41,23 +41,30 @@ public enum ImageRepresentation {
 }
 
 /// Class to interact with the `Files` collection in the backend.
-open class FileStore {
+open class FileStore<FileType: File> {
     
-    public typealias FileCompletionHandler = (File?, Swift.Error?) -> Void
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/FileStore.swift
```diff
@@ -61,20 +61,25 @@ public enum ImageRepresentation {
...
+    public typealias FileArrayCompletionHandler = ([FileType]?, Swift.Error?) -> Void
     
     internal let client: Client
-    internal let cache: FileCache?
+    internal let cache: AnyFileCache<FileType>?
     
     /// Factory method that returns a `FileStore`.
-    open class func getInstance(_ client: Client = sharedClient) -> FileStore {
-        return FileStore(client: client)
+    open class func getInstance<FileType: File>(client: Client = sharedClient) -> FileStore<FileType> {
```
heyzooi: mark this method as deprecated and use the constructor instead

PR #212, opened 2017-05-30
MLIBZ-1833: client ID to loginWithMic call

heyzooi:
#### Description

Client ID added to loginWithMic call

#### Changes

* Optional parameter on every MIC call to specify the client id

#### Tests

* No additional unit tests

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -273,6 +273,7 @@ open class Client: Credential {
         if let user = keychain.user {
             user.client = self
             activeUser = user
+            clientId = keychain.clientId
             let customUser = user as! U
             completionHandler(.success(customUser))
         } else if let kinveyAuth = sharedKeychain?.kinveyAuth {
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -47,6 +47,12 @@ open class Client: Credential {
         }
     }
     
+    internal var clientId: String? {
+        willSet {
+            keychain.clientId = clientId
+        }
+    }
+    
     private var accessGroup: String?
     
     private var keychain: Keychain {
--- a/Kinvey/Kinvey/MIC.swift
+++ b/Kinvey/Kinvey/MIC.swift
@@ -44,15 +44,15 @@ open class MIC {
         return code
     }
     
-    open class func urlForLogin(redirectURI: URL, loginPage: Bool = true, client: Client = sharedClient) -> URL {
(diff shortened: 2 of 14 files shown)


tejasranade:
@heyzooi needs tests, but looks good otherwise.

Kinvey/Kinvey/MIC.swift
```diff
@@ -130,9 +153,13 @@ open class MIC {
...
     @discardableResult
-    class func login<U: User>(refreshToken: String, client: Client = sharedClient, completionHandler: User.UserHandler<U>? = nil) -> Request {
+    class func login<U: User>(
+        refreshToken: String,
+        client: Client = sharedClient,
+        completionHandler: User.UserHandler<U>? = nil
+    ) -> Request {
         let requests = MultiRequest()
-        let request = client.networkRequestFactory.buildOAuthGrantRefreshToken(refreshToken: refreshToken)
+        let request = client.networkRequestFactory.buildOAuthGrantRefreshToken(refreshToken: refreshToken, clientId: nil)
```
heyzooi: @tejasranade question: how should we handle refresh tokens if you are using a `clientId`?

tejasranade: That's a very good question. Can we store the `client_id` along with the rest of the user information?

Kinvey/Kinvey/Client.swift
```diff
@@ -267,6 +273,7 @@ open class Client: Credential {
         if let user = keychain.user {
             user.client = self
             activeUser = user
+            clientId = keychain.clientId
             let customUser = user as! U
             completionHandler(.success(customUser))
         } else if let kinveyAuth = sharedKeychain?.kinveyAuth {
```
heyzooi: @tejasranade should we do something about client_id and SSO too?

tejasranade: I don't completely understand your question. `client_id` should be stored regardless of SSO (whether the app is using SSO or not). If you're thinking about how the `client_id` factors into multiple apps that use SSO, it's a good question to ask @andrewmaillet

heyzooi: right, we are storing the client_id in the keychain, but if the user is using SSO it would be a shared keychain, so my question is if we should allow the user to specify a `client_id` in this case.

tejasranade: My understanding is that the `client_id` in a shared keychain will still correspond to the active user, so we should treat it exactly the same way. This assumes that 2 apps that want to share a MIC config will also share the same `client_id`.

PR #215, opened 2017-06-20
MLIBZ-1908: x-kinvey-custom-request-properties header

heyzooi:
#### Description

Adding `X-Kinvey-Custom-Request-Properties` header if any `client.customRequestProperties` was set

#### Changes

* The property `client.customRequestProperties` already exists but the HTTP request was ignoring it, but now it's considering

#### Tests

* New unit test to check if the header was added to the request

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -92,11 +92,7 @@ open class Client: Credential {
     /// Timeout interval for this client instance.
     open var timeoutInterval: TimeInterval = 60
     
-    /// App version for this client instance.
-    open var clientAppVersion: String?
-    
-    /// Custom request properties for this client instance.
-    open var customRequestProperties: [String : String] = [:]
+    open var options: Options?
     
     /// The default value for `apiHostName` variable.
     open static let defaultApiHostName = URL(string: "https://baas.kinvey.com/")!
@@ -356,7 +352,7 @@ open class Client: Credential {
             }
             return LocalRequest()
         }
-        let request = networkRequestFactory.buildAppDataPing()
+        let request = networkRequestFactory.buildAppDataPing(options: options)
         Promise<EnvironmentInfo> { fulfill, reject in
             request.execute() { data, response, error in
                 if let response = response,
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -92,6 +92,9 @@ open class Client: Credential {
     /// Timeout interval for this client instance.
     open var timeoutInterval: TimeInterval = 60
     
+    /**
+     Hold default optional values for all calls made by this `Client` instance
+     */
     open var options: Options?
     
     /// The default value for `apiHostName` variable.
--- a/Kinvey/Kinvey/CustomEndpoint.swift
+++ b/Kinvey/Kinvey/CustomEndpoint.swift
@@ -54,8 +54,17 @@ open class CustomEndpoint {
     /// Completion handler block for execute custom endpoints.
     public typealias CompletionHandler<T> = (T?, Swift.Error?) -> Void
     
-    private static func callEndpoint(_ name: String, params: Params? = nil, client: Client, completionHandler: DataResponseCompletionHandler? = nil) -> Request {
-        let request = client.networkRequestFactory.buildCustomEndpoint(name)
+    private static func callEndpoint(
+        _ name: String,
+        params: Params? = nil,
+        options: Options?,
+        completionHandler: DataResponseCompletionHandler? = nil
+    ) -> Request {
+        let client = options?.client ?? sharedClient
+        let request = client.networkRequestFactory.buildCustomEndpoint(
+            name,
+            options: options
+        )
         if let params = params {
             request.request.setValue("application/json", forHTTPHeaderField: "Content-Type")
             switch params.value {
@@ -73,11 +82,22 @@ open class CustomEndpoint {
     /// Executes a custom endpoint by name and passing the expected parameters.
     @discardableResult
     @available(*, deprecated, message: "Please use the generic version of execute(params: CustomEndpoint.Params?) method")
-    open static func execute(_ name: String, params: JsonDictionary? = nil, client: Client = sharedClient, completionHandler: CompletionHandler<JsonDictionary>? = nil) -> Request {
+    open static func execute(
+        _ name: String,
+        params: JsonDictionary? = nil,
+        client: Client = sharedClient,
+        completionHandler: CompletionHandler<JsonDictionary>? = nil
+    ) -> Request {
         let params = params != nil ? Params(params!) : nil
         var request: Request!
         Promise<JsonDictionary> { fulfill, reject in
-            request = callEndpoint(name, params: params, client: client) { data, response, error in
+            request = callEndpoint(
+                name,
+                params: params,
+                options: Options(
+                    client: client
+                )
+            ) { data, response, error in
                 if let response = response, response.isOK, let json: JsonDictionary = client.responseParser.parse(data) {
                     fulfill(json)
                 } else {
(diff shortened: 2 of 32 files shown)


Kinvey/Kinvey/Client.swift
```diff
@@ -96,7 +96,7 @@ open class Client: Credential {
     open var clientAppVersion: String?
     
     /// Custom request properties for this client instance.
-    open var customRequestProperties: [String : String] = [:]
+    open var customRequestProperties: [String : Any]?
```
tejasranade: @heyzooi if the property is called `requestConfiguration` in other places, it should be called something similar here as well. Also worth discussing - if individual methods have the requestConfig option, is there any reason to expose a global config? Overriding, precedence etc adds a lot of (potentially unnecessary) complexity.

Kinvey/Kinvey/CustomEndpoint.swift
```diff
@@ -103,11 +134,18 @@ open class CustomEndpoint {
...
+        params: Params? = nil,
+        client: Client = sharedClient,
+        requestConfiguration: RequestConfiguration? = nil,
+        completionHandler: CompletionHandler<JsonDictionary>? = nil
+    ) -> Request {
         return execute(
             name,
             params: params,
-            client: client
+            client: client,
```
tejasranade: @heyzooi just an idea - a majority of methods in the public API have the `client` parameter. What if we make the `client` a part of requestConfig (defaulting to sharedClient)?

Kinvey/Kinvey/DataStore.swift
```diff
@@ -186,8 +186,17 @@ open class DataStore<T: Persistable> where T: NSObject {
      - returns: A `Request` instance which will allow cancel the request later
      */
     @discardableResult
-    open func find(_ id: String, readPolicy: ReadPolicy? = nil, completionHandler: @escaping ObjectCompletionHandler) -> Request {
-        return find(id, readPolicy: readPolicy) { (result: Result<T, Swift.Error>) in
+    open func find(
+        _ id: String,
+        readPolicy: ReadPolicy? = nil,
```
tejasranade: We should deprecate the `readPolicy` parameter and make it part of the `requestConfig`.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -229,8 +249,19 @@ open class DataStore<T: Persistable> where T: NSObject {
      - returns: A `Request` instance which will allow cancel the request later
      */
     @discardableResult
-    open func find(_ query: Query = Query(), deltaSet: Bool? = nil, readPolicy: ReadPolicy? = nil, completionHandler: @escaping ArrayCompletionHandler) -> Request {
-        return find(query, deltaSet: deltaSet, readPolicy: readPolicy) { (result: Result<[T], Swift.Error>) in
+    open func find(
+        _ query: Query = Query(),
+        deltaSet: Bool? = nil,
```
tejasranade: Similar to `readPolicy`, the `deltaSet` parameter can also be in the `requestConfig`.

Kinvey/Kinvey/FileStore.swift
```diff
@@ -98,12 +98,20 @@ open class FileStore<FileType: File> {
     
     /// Uploads a `UIImage` in a PNG or JPEG format.
     @discardableResult
-    open func upload(_ file: FileType, image: NSImage, imageRepresentation: ImageRepresentation = .png, ttl: TTL? = nil, completionHandler: FileCompletionHandler? = nil) -> Request {
+    open func upload(
+        _ file: FileType,
+        image: NSImage,
+        imageRepresentation: ImageRepresentation = .png,
+        ttl: TTL? = nil,
```
tejasranade: TTL can be part of `requestConfig`.

Kinvey/Kinvey/RequestFactory.swift
```diff
@@ -63,3 +63,22 @@ protocol RequestFactory {
     func buildLiveStreamUnsubscribe(streamName: String, userId: String, deviceId: String) -> HttpRequest
     
 }
+
+//Allow to set a per request
+public struct RequestConfiguration {
```
tejasranade: @heyzooi this makes sense for the `customRequestHeaders` problem you're solving. As my other comments suggest, you should consider using  this requestConfig definition to encapsulate other custom parameters that are exposed on each API.

Kinvey/Kinvey/Client.swift
heyzooi: good question... I'm going to check with the requester of this feature if make sense to keep a global version

Kinvey/Kinvey/CustomEndpoint.swift
heyzooi: I think it's a good idea, but it will result in a huge breaking change. Do you think it's worth doing it? I think it's worth it because I'm don't seeing many people using a custom reference for the `client` anyway.

Kinvey/Kinvey/DataStore.swift
heyzooi: I agree, the only catch is `readPolicy` and `writePolicy` are not available for all the requests that requestConfig would be present. We could either subclass `RequestConfiguration` or leave explicit in the docs which options are considered for each call. My opinion: subclass would be better since it make its clear what are the options available.

Kinvey/Kinvey/DataStore.swift
heyzooi: agree, we don't need to decide which approach to do according to the previous comment.

Kinvey/Kinvey/FileStore.swift
heyzooi: totally agree, all those changes will also help to solve this problem here: https://hub.codebeat.co/docs/software-quality-metrics#arity

Kinvey/Kinvey/RequestFactory.swift
heyzooi: agree

Kinvey/Kinvey/DataStore.swift
```diff
@@ -186,8 +186,15 @@ open class DataStore<T: Persistable> where T: NSObject {
      - returns: A `Request` instance which will allow cancel the request later
      */
     @discardableResult
-    open func find(_ id: String, readPolicy: ReadPolicy? = nil, completionHandler: @escaping ObjectCompletionHandler) -> Request {
-        return find(id, readPolicy: readPolicy) { (result: Result<T, Swift.Error>) in
+    open func find(
```
tejasranade: @heyzooi  This should be marked deprecated.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -207,11 +214,44 @@ open class DataStore<T: Persistable> where T: NSObject {
      - returns: A `Request` instance which will allow cancel the request later
      */
     @discardableResult
-    open func find(_ id: String, readPolicy: ReadPolicy? = nil, completionHandler: @escaping (Result<T, Swift.Error>) -> Void) -> Request {
+    open func find(
```
tejasranade: This should also be deprecated.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -229,8 +269,17 @@ open class DataStore<T: Persistable> where T: NSObject {
      - returns: A `Request` instance which will allow cancel the request later
      */
     @discardableResult
-    open func find(_ query: Query = Query(), deltaSet: Bool? = nil, readPolicy: ReadPolicy? = nil, completionHandler: @escaping ArrayCompletionHandler) -> Request {
-        return find(query, deltaSet: deltaSet, readPolicy: readPolicy) { (result: Result<[T], Swift.Error>) in
+    open func find(
```
tejasranade: Also needs deprecation.

Kinvey/Kinvey/RequestFactory.swift
```diff
@@ -10,56 +10,97 @@ import Foundation
...
+    func buildLiveStreamSubscribe(streamName: String, userId: String, deviceId: String, options: Options?) -> HttpRequest
+    func buildLiveStreamUnsubscribe(streamName: String, userId: String, deviceId: String, options: Options?) -> HttpRequest
+    
+}
+
+//Allow to set a per request
+public struct Options {
+    
+    public var client: Client?
+    public var clientId: String?
```
tejasranade: @heyzooi why is this necessary?

Kinvey/Kinvey/RequestFactory.swift
```diff
@@ -10,56 +10,97 @@ import Foundation
...
+    func buildOAuthGrantRefreshToken(refreshToken: String, options: Options?) -> HttpRequest
+    
+    func buildLiveStreamGrantAccess(streamName: String, userId: String, acl: LiveStreamAcl, options: Options?) -> HttpRequest
+    func buildLiveStreamPublish(streamName: String, userId: String, options: Options?) -> HttpRequest
+    func buildLiveStreamSubscribe(streamName: String, userId: String, deviceId: String, options: Options?) -> HttpRequest
+    func buildLiveStreamUnsubscribe(streamName: String, userId: String, deviceId: String, options: Options?) -> HttpRequest
+    
+}
+
+//Allow to set a per request
```
tejasranade: Since options is going to be used for more advanced use cases, I suggest writing more detailed API docs.

Kinvey/Kinvey/RequestFactory.swift
```diff
@@ -10,56 +10,97 @@ import Foundation
...
+    public var writePolicy: WritePolicy?
+    public var timeout: TimeInterval?
+    
+    /// App version for this client instance.
+    public var clientAppVersion: String?
+    
+    /// Custom request properties for this client instance.
+    public var customRequestProperties: [String : Any]?
+    
+    public init(
```
tejasranade: @heyzooi Should there be an `init` with no parameters? Since all properties are public, maybe it makes sense to just `init()` and then set the ones you care about?

Kinvey/Kinvey/RequestFactory.swift
heyzooi: this is for MIC remember?

PR #216, opened 2017-06-21
MLIBZ-1927: query with plus sign

heyzooi:
#### Description

Encoding "+" inside URL queries

#### Changes

* Workaround since `URLQueryItem` does not encode "+" in URL queries

#### Tests

* Unit test that saves and then try to retrieve by query a string value that contains a plug sign, like "C++" for example

### DIFF
--- a/Kinvey/Kinvey/HttpRequest.swift
+++ b/Kinvey/Kinvey/HttpRequest.swift
@@ -358,6 +358,16 @@ internal class HttpRequest: TaskProgressRequest, Request {
         if let clientAppVersion = client.clientAppVersion {
             request.setValue(clientAppVersion, forHTTPHeaderField: KinveyHeaderField.clientAppVersion)
         }
+        if let url = request.url,
+            let query = url.query,
+            query.contains("+"),
+            let decodedQuery = query.removingPercentEncoding,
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/HttpRequest.swift
```diff
@@ -358,6 +358,16 @@ internal class HttpRequest: TaskProgressRequest, Request {
         if let clientAppVersion = client.clientAppVersion {
             request.setValue(clientAppVersion, forHTTPHeaderField: KinveyHeaderField.clientAppVersion)
         }
+        if let url = request.url,
+            let query = url.query,
+            query.contains("+"),
```
vinaygahlawat: @heyzooi Makes sense, but does the same thing have to happen for the other special URL characters:

`$-_.+!*'(),`

heyzooi: those are already handled by the Apple's class `URLQueryItem`, the only one missing is `+` which Apple is being strict to the RFC spec, but W3C recommends to encode. Please look the note section in the link below:

https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems

PR #219, opened 2017-07-07
MLIBZ-1951: crashing with empty body error

heyzooi:
#### Description

Crash was happening when a DLC/BL is returning an empty body error

#### Changes

* Avoiding force unwrap optionals when building error objects

#### Tests

* No need for additional unit test

### DIFF
--- a/Kinvey/Kinvey/Error.swift
+++ b/Kinvey/Kinvey/Error.swift
@@ -179,16 +179,46 @@ public enum Error: Swift.Error, LocalizedError, CustomStringConvertible, CustomD
         return unknownJsonError(httpResponse: httpResponse, data: data, json: json)
     }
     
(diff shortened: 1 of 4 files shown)


ivanstoyanov:
"The diff coverage is 100%" - how does that work, if the commit has no tests?

heyzooi:
@ivanstoyanov our current tests are already covering this part of the code. we were forcing an unwrap of a value before (which could cause a crash). the unwrap is now removed which will not crash anymore.

tejasranade:
I did not see Ivan's comment before my review :) We're asking slightly different things, I think.
IMO it's useful to add a test that mocks the kind of response that resulted in this bug.

heyzooi:
@tejasranade right... doing it now

heyzooi:
@tejasranade unit test is now added. let me know if I can merge now

PR #220, opened 2017-07-08
MLIBZ-1933: Increase documentation coverage

heyzooi:
#### Description

Increase documentation coverage

#### Changes

* Adding basic documentation around the code base

#### Tests

* No changes in the code itself, so no need for additional unit tests

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -344,8 +344,9 @@ open class Client: Credential {
     }
     
     /**
-     Check if the `appKey` and `appSecret` properties are correct doing a ping
-     call to the server.
+     Checks connectivity to your backend. A successful response returns a
+     summary of your backend environment and confirms that the app can talk to
+     the backend.
      */
     @discardableResult
     public func ping(completionHandler: @escaping (Result<EnvironmentInfo, Swift.Error>) -> Void) -> Request {
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -327,6 +327,10 @@ open class Client: Credential {
         return Client.fileURL(appKey: self.appKey!, tag: tag)
     }
     
+    /**
+     Check if the `appKey` and `appSecret` properties are correct doing a ping
+     call to the server.
+     */
     @discardableResult
     public func ping(completionHandler: @escaping (EnvironmentInfo?, Swift.Error?) -> Void) -> Request {
         return ping() { (result: Result<EnvironmentInfo, Swift.Error>) in
@@ -339,6 +343,10 @@ open class Client: Credential {
         }
     }
     
+    /**
+     Check if the `appKey` and `appSecret` properties are correct doing a ping
+     call to the server.
+     */
     @discardableResult
     public func ping(completionHandler: @escaping (Result<EnvironmentInfo, Swift.Error>) -> Void) -> Request {
         guard let _ = appKey, let _ = appSecret else {
@@ -371,11 +379,19 @@ open class Client: Credential {
     }
 }
 
+/// Environment Information for a specific `appKey` and `appSecret`
 public struct EnvironmentInfo: StaticMappable {
     
+    /// Version of the backend
     public let version: String
+    
+    /// Hello message from Kinvey
     public let kinvey: String
+    
+    /// Application Name
     public let appName: String
+    
+    /// Environment Name
     public let environmentName: String
     
     public static func objectForMapping(map: Map) -> BaseMappable? {
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -90,6 +90,15 @@ open class DataStore<T: Persistable> where T: NSObject {
     
     /**
      Deprecated method. Please use `collection()` instead.
+     - Parameters:
+       - type: (Optional) Type for the new DataStore instance. Default value:
+     `.cache`
+       - deltaSet: (Optional) Enables delta set cache. Default value: `nil`
+       - client: (Optional) `Client` instance to be used for all requests.
+     Default value: `sharedClient`
+       - tag: (Optional) Tag the store and separate stores in different tags.
+     Default value: `defaultTag`
+     - Returns: An instance of `DataStore` which can be a new instance or a cached instance if you are passing a `tag` parameter.
      */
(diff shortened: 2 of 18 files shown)


Kinvey/Kinvey/DataStore.swift
```diff
@@ -281,10 +304,14 @@ open class DataStore<T: Persistable> where T: NSObject {
     }
     
     /**
-     Gets a count of how many records that matches with the (optional) query passed by parameter.
-     - parameter query: The query used to filter the results
-     - parameter readPolicy: Enforces a different `ReadPolicy` otherwise use the client's `ReadPolicy`. Default value: `nil`
-     - parameter completionHandler: Completion handler to be called once the respose returns
+     Gets a count of how many records that matches with the (optional) query
```
tejasranade: @heyzooi "count of records that matches with the (optional) query parameter".

Kinvey/Kinvey/DataStore.swift
```diff
@@ -281,10 +304,14 @@ open class DataStore<T: Persistable> where T: NSObject {
...
+     passed by parameter.
+     - parameter query: (Optional) The query used to filter the results. Default
+     value: `nil`
```
tejasranade: When `query` is `nil`, gets the total count of the collection.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -281,10 +304,14 @@ open class DataStore<T: Persistable> where T: NSObject {
...
+     - parameter readPolicy: (Optional) Enforces a different `ReadPolicy`
```
tejasranade: This doesn't quite make sense. If you say "...otherwise uses the client's `ReadPolicy`, then the default is not `nil`. Actually, is it even the client's ReadPolicy? This seems more accurate - by default, the `ReadPolicy` is inferred from the store's type. If provided, the `ReadPolicy` overrides the default set by the store type.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -299,9 +326,38 @@ open class DataStore<T: Persistable> where T: NSObject {
         return request
     }
     
+    /**
+     Custom aggregation function performed in the backend.
```
tejasranade: What do you mean by `performed in the backend`? If this API only works against the backend, you should say that specifically instead.

"Custom aggregation function. Note: this function does not work on local data. It must be run against the backend."

Kinvey/Kinvey/DataStore.swift
```diff
@@ -299,9 +326,38 @@ open class DataStore<T: Persistable> where T: NSObject {
...
+     - parameter keys: (Optional) Property names that should be grouped. Default
+     value: `nil`
+     - parameter initialObject: Sets an initial object that contain initial
```
tejasranade: Typo: "contains"

Kinvey/Kinvey/DataStore.swift
```diff
@@ -911,6 +1182,9 @@ open class DataStore<T: Persistable> where T: NSObject {
         return realtimeRouter
     }
     
+    /**
+     Subscribe and start listening changes in the collection
```
tejasranade: Typo: "listening to changes"

Kinvey/Kinvey/FileStore.swift
```diff
@@ -89,6 +93,10 @@ open class FileStore<FileType: File> {
         return FileStore<FileType>(client: client)
     }
     
+    /**
+     Constructor that takes a specific `Client` instance or use the
```
tejasranade: More useful to describe what it constructs - "Creates a new file cache on disk, at the path specified by `client.fileURL()`.

Kinvey/Kinvey/Kinvey.swift
```diff
@@ -11,9 +11,13 @@ import XCGLogger
 
 let ObjectIdTmpPrefix = "tmp_"
 
-/// Shared client instance for simplicity. Use this instance if *you don't need* to handle with multiple Kinvey environments.
+/**
+ Shared client instance for simplicity. Use this instance if *you don't need* to
```
tejasranade: Better to say something along these lines - 

"The `sharedClient` is set to the default client instance that you initialize with `Kinvey.init`. All methods that use a `client` will default to this instance. If you intend to use multiple backend apps or environments, you should override this default by providing a separate `Client` instance."

Kinvey/Kinvey/Kinvey.swift
```diff
@@ -34,9 +38,36 @@ fileprivate extension Keychain {
     
 }
 
+/**
+ Define how detailed operations should be logged. Here's the asceding order
```
tejasranade: Typo: "ascending"

Kinvey/Kinvey/Realtime.swift
```diff
@@ -10,9 +10,20 @@ import Foundation
 import ObjectMapper
 import PromiseKit
 
+/// Tells the current status for the realtime connection
 public enum RealtimeStatus {
     
-    case connected, disconnected, reconnected, unexpectedDisconnect
+    /// Connection is stablished
```
tejasranade: Typo: "established"

Kinvey/Kinvey/Result.swift
```diff
@@ -8,9 +8,13 @@
 
 import Foundation
 
+/// Enumeration that represents a result expected usually after an async call
```
tejasranade: Is the `expected usually after async call` necessary? How about just `represents the result of an operation / API`

tejasranade: Also, this might be the right place to show how to use this enum. Do you think we should describe that?

Kinvey/Kinvey/Client.swift
```diff
@@ -339,6 +343,10 @@ open class Client: Credential {
         }
     }
     
+    /**
+     Check if the `appKey` and `appSecret` properties are correct doing a ping
```
heyzooi: agree, doing it.

Kinvey/Kinvey/FileStore.swift
heyzooi: `client.fileURL()` is an internal method

Kinvey/Kinvey/Result.swift
heyzooi: definitely

heyzooi:
@tejasranade the changes that you requested are now in place

PR #221, opened 2017-07-14
MLIBZ-1970: bug fix for count not translating queries

heyzooi:
#### Description

Queries must be translated if the property has an alias name
Bug detected during random look at the code

#### Changes

* Fix for count() method

#### Tests

* New unit test to cover the use case

### DIFF
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -329,7 +329,7 @@ open class DataStore<T: Persistable> where T: NSObject {
     @discardableResult
     open func count(_ query: Query? = nil, readPolicy: ReadPolicy? = nil, completionHandler: ((Result<Int, Swift.Error>) -> Void)?) -> Request {
         let readPolicy = readPolicy ?? self.readPolicy
-        let operation = CountOperation<T>(query: query, readPolicy: readPolicy, cache: cache, client: client)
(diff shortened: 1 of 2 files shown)


Kinvey/Kinvey/DataStore.swift
```diff
@@ -329,7 +329,7 @@ open class DataStore<T: Persistable> where T: NSObject {
     @discardableResult
     open func count(_ query: Query? = nil, readPolicy: ReadPolicy? = nil, completionHandler: ((Result<Int, Swift.Error>) -> Void)?) -> Request {
         let readPolicy = readPolicy ?? self.readPolicy
-        let operation = CountOperation<T>(query: query, readPolicy: readPolicy, cache: cache, client: client)
+        let operation = CountOperation<T>(query: Query(query: query ?? Query(), persistableType: T.self), readPolicy: readPolicy, cache: cache, client: client)
```
tejasranade: In general, this seems too verbose. Why do we need a syntax like the following?
`query: Query(query: query?? Query(), ...)`  😕

heyzooi: @tejasranade in order the query be able to translate the properties, the query needs a `persistableType` so we can figure how to translate the query. that's why we need to wrapper the original query into a new "typed" query.

tejasranade: Yeah, but some of this can go a level deeper e.g. the `CountOperation` knows the type `T` and the query, so it could construct this inner query object.

PR #225, opened 2017-08-02
MLIBZ-1969 and MLIBZ-1978: implementing lazy load to increase performance

heyzooi:
#### Description

Implementing lazy load to increase performance

#### Changes

* All `find()` methods now return `AnyRandomAccessCollection` to allow the usage of lazy loading
* Saving data in cache is now done by `dynamicCreate()` instead of `create()`

#### Tests

* Performance Unit Test was added to measure the performance


heyzooi:
@tejasranade this is not ready to merge yet

Kinvey/Kinvey/DataStore.swift
```diff
@@ -317,7 +317,7 @@ open class DataStore<T: Persistable> where T: NSObject {
             query,
             deltaSet: deltaSet,
             readPolicy: readPolicy
-        ) { (result: Result<[T], Swift.Error>) in
+        ) { (result: Result<AnyRandomAccessCollection<T>, Swift.Error>) in
```
tejasranade: @heyzooi isn't this a significant breaking change?

Kinvey/Kinvey/Query.swift
```diff
@@ -70,19 +77,34 @@ public final class Query: NSObject, BuilderType, Mappable {
             }
             return NSExpression(forKeyPath: keyPath)
         case .constantValue:
-            if otherSideExpression.expressionType == .keyPath,
-                let (_, optionalTransform) = persistableType?.propertyMapping(otherSideExpression.keyPath),
-                let transform = optionalTransform
-            {
-                return NSExpression(forConstantValue: transform.transformToJSON(expression.constantValue))
-            }
+            #if !os(watchOS)
```
tejasranade: I'm guessing you had to add this for some compilation error on watchOS. Just out of curiosity, what *should* happen in this code if run on watchOS? Can you add some default behavior or error message that would make sense?

Kinvey/Kinvey/DataStore.swift
heyzooi: yes, it is... a way to avoid this breaking change is use the same pattern that Apple use, which is add a property called `lazy`, so `dataStore.lazy.find()` would have those new methods.

Kinvey/Kinvey/Query.swift
heyzooi: this is because `MKShape` is not available on watchOS

heyzooi:
@tejasranade I think it's now ready to merge, let me know when I can merge

PR #227, opened 2017-08-15
MLIBZ-1934: improving codebeat score

heyzooi:
#### Description
Improving codebeat score

#### Changes
* Changes are based in the Codebeat report

#### Tests
* No changes needed

### DIFF
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -2013,6 +2013,23 @@ open class DataStore<T: Persistable> where T: NSObject {
         return realtimeRouter
     }
     
+    func execute(request: HttpRequest) -> Promise<RealtimeRouter> {
+        return Promise<RealtimeRouter> { fulfill, reject in
+            do {
+                let realtimeRouter = try self.realtimeRouter()
+                request.execute() { (data, response, error) in
+                    if let response = response, response.isOK {
+                        fulfill(realtimeRouter)
+                    } else {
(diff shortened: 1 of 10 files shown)


Kinvey/Kinvey/HttpRequestFactory.swift
```diff
@@ -625,15 +617,19 @@ class HttpRequestFactory: RequestFactory {
     
     // MARK: Realtime
     
-    func buildUserRegisterRealtime(
-        user: User,
+    private func client(options: Options?) -> Client {
```
tejasranade: Seems unnecessary.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -2029,20 +2046,9 @@ open class DataStore<T: Persistable> where T: NSObject {
...
-                        fulfill(realtimeRouter)
-                    } else {
-                        reject(buildError(data, response, error, self.client))
-                    }
-                }
-            } catch {
-                reject(error)
-            }
-        }.then { realtimeRouter in
+        execute(
```
heyzooi: @tejasranade well, `RealtimeRouter` as the name suggest is just a router to map the signals between consumer and producer. The requirement of `subscribe`, `unsubscribe` is out of scope for the router itself.

Kinvey/Kinvey/HttpRequestFactory.swift
heyzooi: hummm, yeah, probably true! I will change that!

PR #231, opened 2017-08-24
MLIBZ-2043: automatic pagination

heyzooi:
#### Description

Automatic pagination (optional, disabled by default) when the result set size exceeds the limit defined by the backend

#### Changes

* `DataStore` additional argument called `autoPagination`
* `FindOperation` checks if additional requests has to be made to get the pagination done

#### Tests

* Unit tests checking the `autoPagination` disabled / enabled and with / without cache


Kinvey/Kinvey/FindOperation.swift
```diff
@@ -192,7 +201,61 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
                             cache.save(entities: entities)
                         }
                     }
-                    completionHandler?(.success(entities))
+                    if self.autoPagination,
+                        entities.count == IntMax(MaxSizePerResultSet)
```
tejasranade: This solution is interesting, but I think it's a little simpler to make the `_count` call first in all cases when `autoPagination` is set. The .NET SDK does it this way.

heyzooi:
@tejasranade it should now be good to merge

heyzooi:
@vinaygahlawat I'm wondering if you could review this one

PR #232, opened 2017-08-25
MLIBZ-2050: LiveStreamAcl default constructor must be available

heyzooi:
#### Description

LiveStreamAcl default constructor was internal, but should be public

#### Changes

* Adding a explicit constructor for LiveStreamAcl

#### Tests

* Unit test added to check if the constructor with no parameters is available

### DIFF
--- a/Kinvey/KinveyTests/RealtimeAclTestCase.swift
+++ b/Kinvey/KinveyTests/RealtimeAclTestCase.swift
@@ -1,0 +1,19 @@
+//
+//  RealtimeAclTestCase.swift
+//  Kinvey
+//
+//  Created by Victor Hugo on 2017-08-25.
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import XCTest
+import Kinvey
+
+class RealtimeAclTestCase: KinveyTestCase {
+    
+    func testLiveStreamAclConsturctorNoParams() {
+        let acl = LiveStreamAcl()
+        XCTAssertNotNil(acl)
+    }
+    
+}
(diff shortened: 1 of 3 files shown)


Kinvey/KinveyTests/RealtimeAclTestCase.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Created by Victor Hugo on 2017-08-25.
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import XCTest
+import Kinvey
+
+class RealtimeAclTestCase: KinveyTestCase {
+    
+    func testLiveStreamAclConsturctorNoParams() {
```
tejasranade: @heyzooi there's a typo in the method name :)
Also, it'll be nice to add some more tests around the other parameters in the constructor.

Kinvey/KinveyTests/RealtimeAclTestCase.swift
```diff
@@ -0,0 +1,19 @@
...
+        let acl = LiveStreamAcl()
+        XCTAssertNotNil(acl)
```
tejasranade: Also, when the constructor is used without parameters, what should be the default values of the properties of the class? Is it worth asserting those?
What about usage of this class? Is that tested anywhere?

heyzooi: no, we don't have tests for ACL for Realtime yet, the only reason for this test is to make sure that a constructor without any parameters is available since it should be the most common use case. Our devcenter has an example that demonstrates this usage: https://devcenter.kinvey.com/ios/guides/live-service#DirectedCommunication

Kinvey/KinveyTests/RealtimeAclTestCase.swift
heyzooi: ok, I will add more tests specifically for ACL on LiveService

heyzooi: as we discussed, I will create new tickets to test ACL on LiveService and merge this PR as is now

PR #233, opened 2017-08-29
MLIBZ-2040: Subclassing File

heyzooi:
#### Description

Subclassing File was not possible because mapping() must be open and not just public

#### Changes

* `File.mapping()` was not `open`

#### Tests

* `MyFile` used in tests is now in its own file so we don't import internal methods from the library causing the unit test to failure catching this bug


heyzooi:
@tejasranade yes, it's coming! @AntonDobrev is just waiting for us to make the release so he can change our DevCenter as well

PR #234, opened 2017-08-29
MLIBZ-1549: _id required, but not _kmd and _acl

heyzooi:
#### Description

Make _id required, but not _kmd and _acl

#### Changes

* `Entity` now validates if `_id` is present

#### Tests

* Unit test to check if is ok if `_kmd` or `_acl` is missing
* Unit test to check if `_id` is missing in a `find(byId:)` throws an error
* Unit test to check if `_id` is missing in a `find()` returns an empty array


heyzooi:
@tejasranade interesting question... I will add a unit test for that as well

heyzooi:
@tejasranade delta set use case is also handled now

PR #237, opened 2017-08-31
MLIBZ-1502: logout async

heyzooi:
#### Description

Make the `logout()` async which will end up invalidating the current auth token

#### Changes

* Adding optional parameters to the `logout()` method

#### Tests

* Unit testing mocking the response from the server

### DIFF
--- a/Kinvey/Kinvey/User.swift
+++ b/Kinvey/Kinvey/User.swift
@@ -996,10 +996,34 @@ open class User: NSObject, Credential, Mappable {
     }
     
     /// Sign out the current active user.
-    open func logout() {
-        if self == client.activeUser {
-            client.activeUser = nil
+    @discardableResult
+    open func logout(
+        options: Options? = nil,
+        completionHandler: ((Result<Void, Swift.Error>) -> Void)? = nil
+    ) -> Request {
+        let request = client.networkRequestFactory.buildUserLogout(
+            user: self,
+            options: options
+        )
+        Promise<Void> { fulfill, reject in
+            request.execute { data, response, error in
+                if let response = response,
+                    response.isOK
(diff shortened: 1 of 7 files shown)


Kinvey/Kinvey/User.swift
```diff
@@ -996,10 +996,34 @@ open class User: NSObject, Credential, Mappable {
...
+            options: options
+        )
+        Promise<Void> { fulfill, reject in
+            request.execute { data, response, error in
+                if let response = response,
+                    response.isOK
+                {
+                    fulfill()
+                } else {
+                    reject(error ?? buildError(data, response, error, self.client))
```
tejasranade: @heyzooi two suggestions -
1 - let's bring this up in standup and check what we're doing on the other platforms for logout failures... does an error in the network operation get reported as an error, or is it still a success (since we logged the user out locally)?
2 - please make sure the docs describe the failure case properly.

Kinvey/KinveyTests/UserTests.swift
```diff
@@ -3865,6 +3865,40 @@ extension UserTests {
         }
     }
     
+    func testLogout() {
```
tejasranade: It'll be useful to test the case where the logout network call fails.

Kinvey/Kinvey/User.swift
heyzooi: @tejasranade as discussed in our standup, the current implementation is the expected behaviour

Kinvey/KinveyTests/UserTests.swift
heyzooi: @tejasranade additional unit test were added

PR #242, opened 2017-09-25
MLIBZ-2087: allow users to retrieve the acl for a live stream

heyzooi:
#### Description

Allow users to retrieve the acl for a live stream

#### Changes

* Add a new method call `streamAccess()`, following a similar name to grant access, which is `grantStreamAccess()`

#### Tests

* Add a check in a existing unit test

### DIFF
--- a/Kinvey/Kinvey/Realtime.swift
+++ b/Kinvey/Kinvey/Realtime.swift
@@ -124,6 +124,38 @@ public class LiveStream<Type: BaseMappable> {
         return request
     }
     
+    /// Grant access to a user for the `LiveStream`
+    @discardableResult
+    public func streamAccess(
+        userId: String,
+        options: Options? = nil,
+        completionHandler: ((Result<LiveStreamAcl, Swift.Error>) -> Void)? = nil
+    ) -> Request {
+        let client = options?.client ?? self.client
+        let request = client.networkRequestFactory.buildLiveStreamAccess(
+            streamName: name,
+            userId: userId,
+            options: options
+        )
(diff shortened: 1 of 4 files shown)


Kinvey/Kinvey/Realtime.swift
```diff
@@ -124,6 +124,38 @@ public class LiveStream<Type: BaseMappable> {
         return request
     }
     
+    /// Grant access to a user for the `LiveStream`
```
tejasranade: As a user of the SDK, how do I retrieve the `acl`? Isn't this class internal?
Also as a side note, the file is still called `Realtime.swift`. Would be nice for it to reflect the new name `LiveService`.

heyzooi: No, `LiveStream` is not internal. The name of the file is `Realtime.swift` but there's a public class in it called `LiveStream` and this class is public. I think i got this `LiveStream` from the original Xamarin implementation

tejasranade: Yeah, I was referring to the file name being `Realtime.swift`. We should keep all the names consistent with our new naming convention (`Live Service`, as opposed to `Realtime` or `LiveStream`).

Kinvey/Kinvey/Realtime.swift
```diff
@@ -124,6 +124,38 @@ public class LiveStream<Type: BaseMappable> {
...
+    @discardableResult
+    public func streamAccess(
```
tejasranade: Can we call this `streamACL`  instead? That's the term used in other SDKs.

PR #243, opened 2017-09-25
MLIBZ-2095: bugfix for the infinite loop if dynamic keyword is not used

heyzooi:
#### Description

Bugfix for the infinite loop in case the user forgets to add the `dynamic` keyword in the property

#### Changes

* Fixing the `ObjCRuntime` class to not get stucked in a infinite loop in case the property does not have a `dynamic` keyword

#### Tests

* Since this is a infinite loop it's hard to find a way to write a unit test for it

### DIFF
--- a/Kinvey/BackgroundFetch/AppDelegate.swift
+++ b/Kinvey/BackgroundFetch/AppDelegate.swift
@@ -12,7 +12,7 @@ import PromiseKit
 
 class Book: Entity {
     
-    var name: String?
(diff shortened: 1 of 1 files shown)


heyzooi:
the fix itself was included during our Swift 4 migration.
Here's the fix: https://github.com/Kinvey/swift-sdk/blob/develop/Kinvey/Kinvey/ObjCRuntime.swift#L86

PR #244, opened 2017-09-26
MLIBZ-2096: allow custom URLSessions

heyzooi:
#### Description

Allow custom URLSessions for background calls for example

#### Changes

* Added a property called `urlSession` in the `Options` class
* Changed all places where a URLSession is used

#### Tests

* No need to add unit tests

### DIFF
--- a/Kinvey/Kinvey/RequestFactory.swift
+++ b/Kinvey/Kinvey/RequestFactory.swift
@@ -71,6 +71,9 @@ public struct Options {
     /// Custom `Client` instance
     public var client: Client?
     
+    /// Custom `URLSession` instance
+    public var urlSession: URLSession?
+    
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/RequestFactory.swift
```diff
@@ -101,6 +104,7 @@ public struct Options {
      */
     public init(
         client: Client? = nil,
+        urlSession: URLSession? = nil,
```
tejasranade: Just a side note, not directly related to this ticket - the constructor of `Options` probably shouldn't expose each property in the class. As the number of properties grows, it'll get too verbose. It's ok to expect developers to set each property individually after they have created an `Options` instance.

PR #246, opened 2017-09-28
MLIBZ-2106: Support for SFAuthenticationSession (new iOS 11 feature)

heyzooi:
#### Description

Support for SFAuthenticationSession (new iOS 11 feature)

#### Changes

* Adding new case for `MICUserInterface` called `safariAuthenticationSession`

#### Tests

* Added option to the test app

### DIFF
--- a/Kinvey/Kinvey/MIC.swift
+++ b/Kinvey/Kinvey/MIC.swift
@@ -35,8 +35,7 @@ open class MIC {
             redirectURI.host?.lowercased() == url.host?.lowercased(),
             let urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: false),
             var queryItems = urlComponents.queryItems
-            else
-        {
+        else {
             return nil
         }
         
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/MIC.swift
```diff
@@ -217,13 +216,19 @@ public enum MICUserInterface {
     
     /// Uses SFSafariViewController
     case safari
+
+    /// Uses SFAuthenticationSession
+    case safariAuthenticationSession
```
tejasranade: @heyzooi is there a reason for not making this the default?

heyzooi: well, not really since if you not running in iOS 11 we just use to the current default, which is the Safari View Controller. do you want me to make this the default?

tejasranade: From my understanding, it seems to make sense to use the authSession all the time (when available).

heyzooi: sure, let me do that and also try to mock some data for unit tests

heyzooi: write unit tests for this is basically impossible now since there's no way to accept the alert displayed by the system. I tried many ways to accomplish that, KIF, XCUIApplication, UIWindow hierarchy, nothing worked!

PR #248, opened 2017-10-13
MLIBZ-2122: Allow data validation upfront

heyzooi:
#### Description

Allow data validation strategies upfront before return the lazy results

#### Changes

* Introduces an optional validation in the `FindOperation`

#### Tests

* Coming soon after we agree with the solution

### DIFF
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -66,6 +66,8 @@ open class DataStore<T: Persistable> where T: NSObject {
     /// DataStoreType defines how the DataStore will behave.
     open let type: StoreType
     
+    open let validationStrategy: ValidationStrategy?
+    
     fileprivate let fileURL: URL?
     
     internal let cache: AnyCache<T>?
@@ -103,7 +105,13 @@ open class DataStore<T: Persistable> where T: NSObject {
      - Returns: An instance of `DataStore` which can be a new instance or a cached instance if you are passing a `tag` parameter.
      */
     @available(*, deprecated: 3.0.22, message: "Please use `collection()` instead")
-    open class func getInstance(_ type: StoreType = .cache, deltaSet: Bool? = nil, client: Client = sharedClient, tag: String = defaultTag) -> DataStore {
+    open class func getInstance(
+        _ type: StoreType = .cache,
+        deltaSet: Bool? = nil,
+        client: Client = sharedClient,
+        tag: String = defaultTag,
+        validationStrategy: ValidationStrategy? = nil
+    ) -> DataStore {
         return collection(type, deltaSet: deltaSet, client: client, tag: tag)
     }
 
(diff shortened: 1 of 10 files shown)


Kinvey/Kinvey/DataStore.swift
```diff
@@ -103,7 +105,13 @@ open class DataStore<T: Persistable> where T: NSObject {
      - Returns: An instance of `DataStore` which can be a new instance or a cached instance if you are passing a `tag` parameter.
      */
     @available(*, deprecated: 3.0.22, message: "Please use `collection()` instead")
-    open class func getInstance(_ type: StoreType = .cache, deltaSet: Bool? = nil, client: Client = sharedClient, tag: String = defaultTag) -> DataStore {
+    open class func getInstance(
```
tejasranade: Needs a reference doc for the new `validationStrategy` parameter.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -115,15 +123,30 @@ open class DataStore<T: Persistable> where T: NSObject {
...
      - returns: An instance of `DataStore` which can be a new instance or a cached instance if you are passing a `tag` parameter.
      */
-    open class func collection(_ type: StoreType = .cache, deltaSet: Bool? = nil, autoPagination: Bool = false, client: Client = sharedClient, tag: String = defaultTag) -> DataStore {
+    open class func collection(
+        _ type: StoreType = .cache,
+        deltaSet: Bool? = nil,
+        autoPagination: Bool = false,
+        client: Client = sharedClient,
+        tag: String = defaultTag,
+        validationStrategy: ValidationStrategy? = nil
```
tejasranade: Same as above. Needs reference doc.

Kinvey/Kinvey/ValidationStrategy.swift
```diff
@@ -0,0 +1,37 @@
...
+//  DataValidationStrategy.swift
+//  Kinvey
+//
+//  Created by Victor Hugo Carvalho Barros on 2017-10-12.
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import Foundation
+
+public enum ValidationStrategy {
```
tejasranade: Needs reference doc.

Kinvey/Kinvey/ValidationStrategy.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    /// Percentagem between 0.0 and 1.0
+    case randomItems(percentage: Double)
```
tejasranade: Can we name this `randomSample`? Usually "sampling" is the term used for what we're doing.

Kinvey/Kinvey/ValidationStrategy.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+//  Created by Victor Hugo Carvalho Barros on 2017-10-12.
+//  Copyright © 2017 Kinvey. All rights reserved.
+//
+
+import Foundation
+
+public enum ValidationStrategy {
+    
+    /// Percentagem between 0.0 and 1.0
```
tejasranade: Typo.
Also, it'll help to define this better. I suggest saying -
"Percentage between 0.0 and 1.0. This number specifies the fraction of entities in a backend response that are validated. Validating a higher percentage of entities upfront results in a performance penalty."

Kinvey/Kinvey/DataStore.swift
heyzooi: done

Kinvey/Kinvey/DataStore.swift
heyzooi: done

PR #250, opened 2017-10-18
MLIBZ-2111: bugfix transforms were being ignored when saving object using dynamic Realm

heyzooi:
#### Description

MLIBZ-2111: bugfix transforms were being ignored when saving object using dynamic Realm

#### Changes

* `RealmCache.save()` using dynamic Realm was not considering cases where a transform needs to be applied

#### Tests

* New unit test added to test a scenario where a Date, which requires some transformation to JSON, is being saved to Realm

### DIFF
--- a/Kinvey/KinveyTests/DataTypeTestCase.swift
+++ b/Kinvey/KinveyTests/DataTypeTestCase.swift
@@ -186,6 +186,60 @@ class DataTypeTestCase: StoreTestCase {
 
     }
     
+    func testDatePull() {
+        signUp()
+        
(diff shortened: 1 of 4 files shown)


Kinvey/KinveyTests/DataTypeTestCase.swift
```diff
@@ -186,6 +186,60 @@ class DataTypeTestCase: StoreTestCase {
 
     }
     
+    func testDatePull() {
```
tejasranade: I think instead of a test like `testDatePull()`, you should write a test that does basic operations on all our supported datatypes. This can go in a future PR.

PR #253, opened 2017-11-06
MLIBZ-2149: DeltaSet Cache

heyzooi:
#### Description

Using a custom endpoint + BL hooks for now to prototype DeltaSet

#### Changes

* `FindOperation`
* DeltaSet on by default now

#### Tests

* Simple unit test for now to help test some situations. More unit tests to come.

### DIFF
--- a/Kinvey/Kinvey/Cache.swift
+++ b/Kinvey/Kinvey/Cache.swift
@@ -11,6 +11,7 @@ import Foundation
 internal protocol CacheType: class {
     
     var ttl: TimeInterval? { get set }
+    var lastPull: Date? { get set }
     
     associatedtype `Type`: Persistable
     
@@ -84,6 +85,15 @@ class AnyCache<T: Persistable>: CacheType {
         }
     }
     
+    var lastPull: Date? {
+        get {
+            return _getLastPull()
+        }
+        set {
+            _setLastPull(newValue)
+        }
+    }
+    
     var dynamic: DynamicCacheType? {
         return _getDynamic()
     }
@@ -91,6 +101,8 @@ class AnyCache<T: Persistable>: CacheType {
     private let _getDynamic: () -> DynamicCacheType?
     private let _getTTL: () -> TimeInterval?
     private let _setTTL: (TimeInterval?) -> Void
+    private let _getLastPull: () -> Date?
+    private let _setLastPull: (Date?) -> Void
     private let _saveEntity: (T) -> Void
     private let _saveEntities: (AnyRandomAccessCollection<Type>) -> Void
     private let _findById: (String) -> T?
@@ -112,6 +124,8 @@ class AnyCache<T: Persistable>: CacheType {
         _getDynamic = { return cache.dynamic }
         _getTTL = { return cache.ttl }
         _setTTL = { cache.ttl = $0 }
+        _getLastPull = { return cache.lastPull }
+        _setLastPull = { cache.lastPull = $0 }
         _saveEntity = cache.save(entity:)
         _saveEntities = cache.save(entities:)
         _findById = cache.find(byId:)
--- a/Kinvey/Kinvey/DataStore.swift
+++ b/Kinvey/Kinvey/DataStore.swift
@@ -349,7 +349,7 @@ open class DataStore<T: Persistable> where T: NSObject {
     @discardableResult
     open func find(
         _ query: Query = Query(),
-        deltaSet: Bool = true,
+        deltaSet: Bool? = nil,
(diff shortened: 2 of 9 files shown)


Kinvey/Kinvey/FindOperation.swift
```diff
@@ -147,145 +147,30 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
...
-                                    reject(error)
-                                }
-                            }
-                        } else {
+            if deltaSet, let cache = self.cache {
+                let lastPull = cache.lastPull ?? Date(timeIntervalSince1970: 0)
+                let request = CustomEndpoint.execute(
+                    "DeltaSet",
+                    params: CustomEndpoint.Params([
+                        "collection" : "Book",
```
tejasranade: This can't be hardcoded. It should be `this.collection`.

Kinvey/Kinvey/FindOperation.swift
```diff
@@ -147,145 +147,30 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
...
+                        "lmt" : lastPull.toString()
+                    ]),
+                    options: Options(
+                        client: client
+                    )
+                ) { (result: Result<JsonDictionary, Swift.Error>) in
+                    switch result {
+                    case .success(let results):
+                        let dateFormatter = DateFormatter()
```
tejasranade: Why not use our previous defined `KinveyDateFormatter` instead?

Kinvey/Kinvey/FindOperation.swift
```diff
@@ -294,8 +179,32 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
                                     reject(error)
                                 }
                             }
+                        } else {
+                            reject(Error.invalidResponse(httpResponse: nil, data: nil))
                         }
-                    } else {
+                    case .failure(let error):
```
tejasranade: In case of failure, we need to decide if the SDK should initiate a regular `fetch` automatically, or whether we should throw an error to the developer and leave it up to them to do it.

Kinvey/Kinvey/DataStore.swift
```diff
@@ -108,7 +108,7 @@ open class DataStore<T: Persistable> where T: NSObject {
     @available(*, deprecated: 3.0.22, message: "Please use `collection()` instead")
     open class func getInstance(
         _ type: StoreType = .cache,
-        deltaSet: Bool? = nil,
+        deltaSet: Bool = true,
```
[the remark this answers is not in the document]
heyzooi: this is the factory method, but you are right about the `Options`, i will make this change

Kinvey/Kinvey/FindOperation.swift
heyzooi: oh sorry, fixing it

Kinvey/Kinvey/FindOperation.swift
heyzooi: right, forgot about that... doing that as well

Kinvey/Kinvey/FindOperation.swift
heyzooi: i don't think we should do a regular fetch in case of failure, this can lead users to assume the delta set is working when is actually not. If delta set is enable i think we should use delta set and inform the user in case of failure

Kinvey/Kinvey/Cache.swift
```diff
@@ -11,6 +11,7 @@ import Foundation
 internal protocol CacheType: class {
     
     var ttl: TimeInterval? { get set }
+    var lastPull: Date? { get set }
```
tejasranade: The difficulty with this API is if the developer uses queries. We need to maintain `lastPull` by query.

Kinvey/Kinvey/FindOperation.swift
```diff
@@ -147,145 +147,29 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
...
-                                            reject(error)
-                                        }
-                                    }
-                                case .failure(let error):
-                                    reject(error)
-                                }
-                            }
-                        } else {
+            if deltaSet, let cache = self.cache {
+                let lastPull = cache.lastPull ?? Date(timeIntervalSince1970: 0)
```
tejasranade: If `cache.lastPull` isn't available, we shouldn't use `Date(timeIntervalSince1970:0)`. We should fallback to regular fetch.

tejasranade:
Since this was shared as a POC, closing the PR.

PR #259, opened 2018-01-11
MLIBZ-2304: making the page size adjustable for auto pagination

heyzooi:
#### Description

Be able to change the default value of 10k for the max page size

#### Changes

* Added new property for `Options`
* `FindOperation` now checks if a value was provided, otherwise uses the default 10k value

#### Tests

* Unit test adjusted to cover the use case


Kinvey/Kinvey/FindOperation.swift
```diff
@@ -102,17 +102,18 @@ internal class FindOperation<T: Persistable>: ReadOperation<T, AnyRandomAccessCo
     
     private func fetchAutoPagination(multiRequest: MultiRequest, count: Int) -> Promise<AnyRandomAccessCollection<T>> {
         return Promise<AnyRandomAccessCollection<T>> { fulfill, reject in
-            let nPages = Int64(ceil(Double(count) / Double(MaxSizePerResultSet)))
+            let maxSizePerResultSet = options?.maxSizePerResultSet ?? MaxSizePerResultSet
```
tejasranade: Corner case, but does zero cause a problem?

Kinvey/KinveyTests/NetworkStoreTests.swift
```diff
@@ -2512,10 +2520,10 @@ class NetworkStoreTests: StoreTestCase {
         
         weak var expectationFind = expectation(description: "Find")
         
-        store.find(options: Options(readPolicy: .forceNetwork)) { (result: Result<AnyRandomAccessCollection<Products>, Swift.Error>) in
+        store.find(options: Options(readPolicy: .forceNetwork, maxSizePerResultSet: pageSizeLimit)) { (result: Result<AnyRandomAccessCollection<Products>, Swift.Error>) in
```
tejasranade: I'm confused, where is the code that adds this to `Options`? What is the relation between `Options` and `RequestFactory`?

heyzooi: the `Options` struct is in the same file of `RequestFactory`. We can create a separate file for `Options` as well. I will do that.

Kinvey/Kinvey/FindOperation.swift
heyzooi: good question, also negative values should not be allowed. I will add some protection to avoid this scenario

Kinvey/Kinvey/Options.swift
```diff
@@ -0,0 +1,83 @@
...
+    public var clientAppVersion: String?
+    
+    /// Custom request properties for this client instance.
+    public var customRequestProperties: [String : Any]?
+    
+    /// Maximum size per result set coming from the backend. Default to 10k records.
+    public var maxSizePerResultSet: Int? {
+        willSet {
+            if let newValue = newValue, newValue <= 0 {
+                fatalError("maxSizePerResultSet must be greater than 0 (zero)")
```
heyzooi: validation to avoid negative values or zero values

PR #264, opened 2018-02-06
MLIBZ-2343: adding Options for Client

heyzooi:
#### Description

Adding Options for Client constructor

#### Changes

- Adding optional parameter in an existing constructor

#### Tests

- Simple addition, no need for unit test

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -238,10 +238,24 @@ open class Client: Credential {
     }
     
     /// Initialize a `Client` instance with all the needed parameters.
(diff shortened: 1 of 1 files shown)


Kinvey/Kinvey/Client.swift
```diff
@@ -238,10 +256,24 @@ open class Client: Credential {
...
+        encryptionKey: Data? = nil,
+        schema: Schema? = nil,
+        options: Options? = nil,
+        completionHandler: @escaping (Result<U?, Swift.Error>) -> Void
+    ) {
         validateInitialize(appKey: appKey, appSecret: appSecret)
         self.encryptionKey = encryptionKey
         self.schemaVersion = schema?.version ?? 0
+        self.options = options
+        if let timeout = options?.timeout {
```
tejasranade: If `options.timeout` is our preferred way to set global timeouts, let's deprecate `Client.timeoutInterval`.

heyzooi: sure, let me do this

PR #265, opened 2018-02-14
MLIBZ-2210: handling user lockdown

heyzooi:
#### Description

Handling a user lockdown case by automatically logout the user if a 401 response is returned from the server

#### Changes

- Checking if a 401 response is returned and then logout the user

#### Tests

- Unit test to mock the situation


heyzooi:
sure, we can do that. I will try that

PR #266, opened 2018-02-14
MLIBZ-2359: FileStore proper result types for sync wait calls

heyzooi:
#### Description

FileStore proper result types for sync wait calls

#### Changes

- Setting the result for each request returned

#### Tests

- Made some additions to existing unit tests

### DIFF
--- a/Kinvey/Kinvey/FileStore.swift
+++ b/Kinvey/Kinvey/FileStore.swift
@@ -112,7 +112,7 @@ open class FileStore<FileType: File> {
         imageRepresentation: ImageRepresentation = .png,
         ttl: TTL? = nil,
         completionHandler: FileCompletionHandler? = nil
-    ) -> AnyRequest<Any> {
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/FileStore.swift
```diff
@@ -112,7 +112,7 @@ open class FileStore<FileType: File> {
         imageRepresentation: ImageRepresentation = .png,
         ttl: TTL? = nil,
         completionHandler: FileCompletionHandler? = nil
-    ) -> AnyRequest<Any> {
+    ) -> AnyRequest<Result<FileType, Swift.Error>> {
```
tejasranade: Needs a doc change? If so, please open a devcenter PR.

heyzooi: we don't have docs for sync / wait calls yet. I will create a ticket for that.

PR #267, opened 2018-02-27
MLIBZ-2337: adding instanceID implementation

heyzooi:
#### Description

Adding instanceID implementation

#### Changes

- Overloading `Client` constructor and `initialize()` method

#### Tests

- Unit test to check if the URLs were changed

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -146,6 +146,29 @@ open class Client: Credential {
     public convenience init<U: User>(
         appKey: String,
         appSecret: String,
+        instanceId: String,
+        accessGroup: String? = nil,
+        schema: Schema? = nil,
+        options: Options? = nil,
+        completionHandler: ((Result<U?, Swift.Error>) -> Void)? = nil
+    ) {
+        self.init()
+        initialize(
+            appKey: appKey,
+            appSecret: appSecret,
+            instanceId: instanceId,
+            accessGroup: accessGroup,
+            schema: schema,
+            options: options
+        ) { (result: Result<U?, Swift.Error>) in
+            completionHandler?(result)
+        }
+    }
+    
+    /// Constructor that already initialize the client. The `initialize` method is called automatically.
+    public convenience init<U: User>(
+        appKey: String,
+        appSecret: String,
         accessGroup: String? = nil,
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/Client.swift
```diff
@@ -256,6 +279,30 @@ open class Client: Credential {
         }
     }
     
+    /// Initialize a `Client` instance with all the needed parameters.
```
tejasranade: This note should be more descriptive. Since this is now our recommended `init()` method, all the parameters should be explained in the reference doc.

Kinvey/Kinvey/Client.swift
```diff
@@ -256,6 +279,30 @@ open class Client: Credential {
...
+        appKey: String,
+        appSecret: String,
+        instanceId: String,
+        accessGroup: String? = nil,
+        encryptionKey: Data? = nil,
+        schema: Schema? = nil,
+        options: Options? = nil,
+        completionHandler: @escaping (Result<U?, Swift.Error>) -> Void
+    ) {
+        return initialize(
```
tejasranade: We should validate the `instanceId` parameter before using it? You are force-unwrapping the `URL()`, so if there are any invalid characters, the method will exception out. Maybe it's better to detect that and throw our own error instead.

heyzooi: agree. done

Kinvey/Kinvey/Client.swift
heyzooi: docs for parameters added

Kinvey/Kinvey/Client.swift
```diff
@@ -256,6 +279,51 @@ open class Client: Credential {
         }
     }
     
+    /**
+     Initialize a `Client` instance.
+     - parameters:
+       - appKey: `App Key` value from Kinvey Console
+       - appSecret: `App Secret` value from Kinvey Console
+       - instanceId: Prefix value of `Host URL` from Kinvey Console. eg: my-instance if the Host URL looks like my-instance.kinvey.com
```
tejasranade: the Host URL looks like `my-instance-baas.kinvey.com`, I think.

heyzooi: then the code itself is wrong, let me double check with our infra guys

PR #268, opened 2018-03-06
MLIBZ-2396: bugfix for memory consumption

heyzooi:
#### Description

Bugfix for memory consumption during sync/find/pull

#### Changes

- Discarding Sync/Wait result for auto-pagination requests

#### Tests

- Unit Test added and compare memory before and after a `pull()`

### DIFF
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -35,6 +35,7 @@ open class Client: Credential {
                 if let sharedKeychain = sharedKeychain {
                     try! sharedKeychain.removeAll()
                 }
+                dataStoreInstances.removeAll()
             } else {
                 if let sharedKeychain = sharedKeychain {
                     try! sharedKeychain.removeAll()
@@ -122,6 +123,8 @@ open class Client: Credential {
     
     ///Default Value for DataStore tag
     open static let defaultTag = Kinvey.defaultTag
+    
+    var dataStoreInstances = [DataStoreTypeTag : AnyObject]()
     
     /// Enables logging for any network calls.
     open var logNetworkEnabled = false
--- a/Kinvey/Kinvey/Client.swift
+++ b/Kinvey/Kinvey/Client.swift
@@ -35,7 +35,6 @@ open class Client: Credential {
                 if let sharedKeychain = sharedKeychain {
                     try! sharedKeychain.removeAll()
                 }
-                dataStoreInstances.removeAll()
             } else {
                 if let sharedKeychain = sharedKeychain {
                     try! sharedKeychain.removeAll()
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/PullOperation.swift
```diff
@@ -19,6 +19,7 @@ internal class PullOperation<T: Persistable>: FindOperation<T> where T: NSObject
         validationStrategy: ValidationStrategy?,
         cache: AnyCache<T>?,
         options: Options?,
+        mustSetRequestResult: Bool = true,
```
vinaygahlawat: Under what circumstances is `mustSetRequestResult` set to false?

Kinvey/KinveyTests/SyncStoreTests.swift
```diff
@@ -2785,4 +2785,67 @@ class SyncStoreTests: StoreTestCase {
...
+            switch $0 {
+            case .success(let results):
+                XCTAssertEqual(results.count, Int64(size))
+            case .failure(let error):
+                XCTFail(error.localizedDescription)
+            }
+            
+            if let startMemory = startMemory, let endMemory = reportMemory() {
+                let diffMemory = endMemory - startMemory
+                XCTAssertLessThan(diffMemory, 300_000_000)
```
vinaygahlawat: Where does the comparison against 300,000,000 come from?

Kinvey/Kinvey/DataStore.swift
```diff
@@ -9,28 +9,6 @@
 import Foundation
 import PromiseKit
 
-class DataStoreTypeTag: Hashable {
```
vinaygahlawat: Was the use of `DataStoreTypeTag` a cause of the memory leak?

Kinvey/Kinvey/Client.swift
```diff
@@ -91,7 +90,17 @@ open class Client: Credential {
...
+        get {
+            return options?.timeout ?? Client.urlSessionConfiguration.timeoutIntervalForRequest
+        }
+        set {
+            if options == nil {
+                options = Options()
+            }
+            options?.timeout = newValue
+        }
+    }
```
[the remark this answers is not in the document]
heyzooi: none, i should create a separate PR for that. Doing that now.

Kinvey/KinveyTests/SyncStoreTests.swift
heyzooi: basically during my tests, the consumption was around `350 MB` without the fix and `250 MB` with the fix, that's why I pick `300 MB` as a threshold

Kinvey/Kinvey/DataStore.swift
heyzooi: not really, again, i should create a separate PR for that. Doing that now.

Kinvey/Kinvey/PullOperation.swift
heyzooi: for auto pagination requests, if you set the result for each request you end up holding all instances in memory, that's why the memory consumption was going to the roof, so in this case we should `false` for this. This is all internal code, so no need to the user to set this.

heyzooi:
@tejasranade and @vinaygahlawat, the unit tests are failing for now until we merge and rebase https://github.com/Kinvey/swift-sdk/pull/272

PR #270, opened 2018-03-08
MLIBZ-1568: adding Swift version to User-Agent HTTP header

heyzooi:
#### Description

Adding Swift version to User-Agent HTTP header

#### Changes

- Changing all HTTP requests to add the Swift version in the `User-Agent` header

#### Tests

- Unit test changed to check if the Swift version is present

### DIFF
--- a/Kinvey/Kinvey/Kinvey.swift
+++ b/Kinvey/Kinvey/Kinvey.swift
@@ -117,6 +117,30 @@ public var logLevel: LogLevel = log.outputLevel.logLevel {
 public let defaultTag = "kinvey"
 let groupId = "_group_"
 
+#if swift(>=5)
+    let swiftVersion = "5 or above"
+#elseif swift(>=4.0.3)
+    let swiftVersion = "4.0.3"
+#elseif swift(>=4.0.2)
+    let swiftVersion = "4.0.2"
+#elseif swift(>=4.0)
+    let swiftVersion = "4.0"
+#elseif swift(>=3.1.1)
+    let swiftVersion = "3.1.1"
+#elseif swift(>=3.1)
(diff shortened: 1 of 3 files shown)


Kinvey/Kinvey/Kinvey.swift
```diff
@@ -117,6 +117,30 @@ public var logLevel: LogLevel = log.outputLevel.logLevel {
...
+#elseif swift(>=3.0.1)
+    let swiftVersion = "3.0.1"
+#elseif swift(>=3.0)
+    let swiftVersion = "3.0"
+#elseif swift(>=2.2.1)
+    let swiftVersion = "2.2.1"
+#elseif swift(>=2.2)
+    let swiftVersion = "2.2"
+#endif
+
```
vinaygahlawat: 1) Why are we doing a `>=` comparison instead of a `==`? Are there some swift versions that are not captured in this statement?

2) What happens when a new version of swift comes out? Will the string still be "5 and above", or will the string have to change?

heyzooi: @vinaygahlawat you are right it should be `==`, but the Swift compiler currently does not support this kind of statement. If you try to use `swift(==2.2)` for example it complains saying `Unexpected platform condition argument: expected a unary comparison, such as '>=2.2'`. That's why I did this chain of multiple `if/elseif`. I also tried to find some runtime call that i could use, but there's no such thing yet in the Swift runtime.

heyzooi: if a new version of swift comes out we need to add one more `elseif`, that's the only way until the swift compiler adds a runtime call that we could use

PR #273, opened 2018-03-15
MLIBZ-2411: send scope param to all mic versions

heyzooi:
#### Description

Send `scope` param to all MIC versions

#### Changes

- Removing the `if` statement and send to all versions

#### Tests

- TBD


heyzooi:
@tejasranade do you think we need a unit test for this? if so, how do you think should we test this scenario?

heyzooi:
@vinaygahlawat `v1` and `v2` just ignore the `scope` values

PR #279, opened 2018-04-09
MLIBZ-2439: User and Entity classes now conforms to Hashable and Equatable

heyzooi:
#### Description

`User` and `Entity` classes now conforms to Hashable and Equatable

#### Changes

- Adding methods for Hashable and Equatable in the `User` and `Entity` class

#### Tests

- Unit tests to double check Hashable and Equatable

### DIFF
--- a/Kinvey/KinveyTests/UserTests.swift
+++ b/Kinvey/KinveyTests/UserTests.swift
@@ -4183,6 +4183,25 @@ extension UserTests {
         userLockDown(mustIncludeSocialIdentity: true)
     }
     
+    func testUserHashable() {
+        let userId = UUID().uuidString
+        let user1 = User(userId: userId)
+        let user2 = User(userId: userId)
+        XCTAssertEqual(user1.hash, user2.hash)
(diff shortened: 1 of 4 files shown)


Kinvey/KinveyTests/UserTests.swift
```diff
@@ -4183,6 +4183,25 @@ extension UserTests {
...
+        let userId = UUID().uuidString
+        let user1 = User(userId: userId)
+        let user2 = User(userId: userId)
+        let set = Set<User>([user1])
+        XCTAssertEqual(user1, user2)
+        XCTAssertTrue(user1 == user2)
+        XCTAssertTrue(user1.isEqual(user2))
+        XCTAssertTrue(set.contains(user2))
+    }
+
```
vinaygahlawat: Does it make sense here to add another test to assert false if the users have different hash values?

heyzooi: sure! I will do that, thanks for the tip

PR #282, opened 2018-04-27
MLIBZ-2465: X-Kinvey-Device-Info header

heyzooi:
#### Description
`X-Kinvey-Device-Info` header using JSON format

#### Changes

- `X-Kinvey-Device-Info` replaces the old `X-Kinvey-Device-Information`

#### Tests

- Unit test to check if the header is being sent and checking if the properties are present

### DIFF
--- a/Kinvey/Kinvey/DeviceInfo.swift
+++ b/Kinvey/Kinvey/DeviceInfo.swift
@@ -1,0 +1,205 @@
+//
+//  DeviceInfo.swift
+//  Kinvey
+//
+//  Created by Victor Hugo Carvalho Barros on 2018-04-26.
+//  Copyright © 2018 Kinvey. All rights reserved.
+//
+
+import Foundation
+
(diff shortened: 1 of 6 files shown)


Kinvey/Kinvey/DeviceInfo.swift
```diff
@@ -0,0 +1,205 @@
...
+    static var platform: String {
+        #if os(iOS)
+            return "iOS"
+        #elseif os(tvOS)
+            return "tvOS"
+        #elseif os(watchOS)
+            return "watchOS"
+        #elseif os(macOS)
+            return "macOS"
+        #elseif os(Linux)
```
thomasconner: Would the platform ever be Linux?

heyzooi: well, Swift can run in Linux, but currently we are not supporting Linux for now. Maybe we can add this to the roadmap.

PR #283, opened 2018-04-27
MLIBZ-2475: default log level must be warning

heyzooi:
#### Description

Default log level must be `.warning`

#### Changes

- `.warning` log level is set when the log is created

#### Tests

- Unit test changed to double check that


thomasconner:
I looked up the associated ticket on JIRA but it didn't provide any information as to why this change was needed. Why was this decided to be changed?

heyzooi:
@thomasconner this was detected when I writing a sample app. Let me add some context in the JIRA ticket.

PR #286, opened 2018-05-08
MLIBZ-2485: skip and limit should not save the query last sync

heyzooi:
#### Description

- Making sure to not save the query last sync date if skip and limit is being used

#### Changes

- `FindOperation` (which is also the base class for `PullOperation`) is now checking for `skip` and `limit` usage, but also considering other scenarios like `autoPagination` for example

#### Tests

- Unit test added to check the cache data

### DIFF
--- a/Kinvey/KinveyTests/DeltaSetCacheTestCase.swift
+++ b/Kinvey/KinveyTests/DeltaSetCacheTestCase.swift
@@ -2203,6 +2203,79 @@ class DeltaSetCacheTestCase: KinveyTestCase {
         }
     }
     
+    func testDeltaSetQuerySkipLimit2ndRequestWithoutSkipAndLimit() {
+        mockResponse { request in
+            switch request.url!.path {
+            case "/appdata/_kid_/Person":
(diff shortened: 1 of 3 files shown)


Kinvey/KinveyTests/DeltaSetCacheTestCase.swift
```diff
@@ -2203,6 +2203,79 @@ class DeltaSetCacheTestCase: KinveyTestCase {
...
+                case .failure(let error):
+                    XCTFail(error.localizedDescription)
+                }
+                expectationPull?.fulfill()
+            }
+            
+            waitForExpectations(timeout: defaultTimeout) { (error) in
+                expectationPull = nil
+            }
+        }
```
vinaygahlawat: Should there be an `XCTAssertNotNil` here, to ensure that there is now an entry in the `_QueryCache` table?

heyzooi: @vinaygahlawat good catch

PR #290, opened 2018-05-17
MLIBZ-2253: oauth resource owner password grant

heyzooi:
#### Description

OAuth Resource Owner password grant flow

https://devcenter.kinvey.com/rest/guides/mobile-identity-connect#resource-owner-grant

Basically it simplifies the steps to login using MIC using username/password

#### Changes

- `MIC` and `User` classes have new methods. Old methods handling the old headless way of login are now deprecated

#### Tests

- Unit Test added

### DIFF
--- a/Kinvey/Kinvey/User.swift
+++ b/Kinvey/Kinvey/User.swift
@@ -402,37 +402,6 @@ open class User: NSObject, Credential, Mappable {
         )
     }
     
-    /// Sign in a user and set as a current active user.
-    @discardableResult
-    open class func login<U: User>(
-        username: String,
-        password: String,
-        options: Options? = nil,
-        completionHandler: ((Result<U, Swift.Error>) -> Void)? = nil
-    ) -> AnyRequest<Result<U, Swift.Error>> {
-        let client = options?.client ?? sharedClient
-        if let error = client.validate() {
-            DispatchQueue.main.async {
-                completionHandler?(.failure(error))
-            }
-            return AnyRequest(LocalRequest<Result<U, Swift.Error>>())
-        }
-
-        let request = client.networkRequestFactory.buildUserLogin(
-            username: username,
-            password: password,
-            options: options,
-            resultType: Result<U, Swift.Error>.self
-        )
-        login(
-            request: request,
-            client: client,
-            userType: U.self,
-            completionHandler: completionHandler
-        )
-        return AnyRequest(request)
-    }
-    
     /**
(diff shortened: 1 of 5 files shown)


Kinvey/Kinvey/User.swift
```diff
@@ -1297,6 +1300,35 @@ open class User: NSObject, Credential, Mappable {
...
+    @discardableResult
+    open class func login<U: User>(
+        username: String,
+        password: String,
+        provider: AuthProvider,
+        options: Options? = nil,
+        completionHandler: ((Result<U, Swift.Error>) -> Void)? = nil
+    ) -> AnyRequest<Result<U, Swift.Error>> {
+        switch provider {
+        case .kinvey:
```
thomasconner: Is this the default `AuthProvider` if none is provided?

heyzooi: no, actually this was the best way that I find to allow overload the `login()` method without need to name it like `loginMIC()`, which in Swift seems weird. So the user actually have to specify `.mic` or `.kinvey` otherwise the compiler would not know if you are trying to login with your Kinvey username/password or with your MIC username/password

thomasconner: I think this is a great solution to prevent having `loginMIC()`. However, it looks like that a user will have to specify a `provider` which will break existing apps. Can't we just specify `AuthProvider.kinvey` as the default? Something like this:

```swift
open class func login<U: User> (
  username: String,
  password: String,
  provider: AuthProvider? = .kinvey,
  completionHandler: ((Result<U, Swift.Error>) -> Void)? = nil
) -> AnyRequest<Result<U, Swift.Error>> {
 //...
}
```

_Not sure my syntax is correct but something like that_

heyzooi: I tried that too, but the problem is if you specify a default value there's no way for the compiler to know which `login()` overload are you referring to. We are also deprecating the old way, so no breaking changes. Let me try if I can eliminate the previous kinvey `login()` method and replace for the new method with the `.kinvey` as default value and see if it will not cause any breaking change. going to test this now.

heyzooi: @thomasconner i was able to move the previous implementation to be inside the new method and set the default value, so no breaking changes and also keep a cleaner code. thanks for your suggestion!

heyzooi: @thomasconner please take another look, it should be good now. thanks for your suggestion!

PR #291, opened 2018-05-18
MLIBZ-2510: removing compiler warnings

heyzooi:
Removing compiler warnings and ignoring deprecation warnings for unit tests

#### Description

- After Swift 4.1 changes the compiler was throwing some warnings

#### Changes

- Mainly `Persistable.swift` and Unit Tests

#### Tests

- No need for additional unit tests

### DIFF
--- a/Kinvey/Kinvey/Geolocation.swift
+++ b/Kinvey/Kinvey/Geolocation.swift
@@ -108,6 +108,14 @@ public func <- (left: inout GeoPoint?, right: (String, Map)) {
     left <- (map, transform)
 }
 
+/// Override operator used during the `propertyMapping(_:)` method.
+public func <- (left: inout GeoPoint!, right: (String, Map)) {
+    let (right, map) = right
+    let transform = GeoPointTransform()
+    kinveyMappingType(left: right, right: map.currentKey!, transform: transform)
+    left <- (map, transform)
+}
+
 func ==(lhs: CLLocationCoordinate2D, rhs: CLLocationCoordinate2D) -> Bool {
     return lhs.latitude == rhs.latitude && lhs.longitude == rhs.longitude
 }
--- a/Kinvey/Kinvey/Geolocation.swift
+++ b/Kinvey/Kinvey/Geolocation.swift
@@ -108,14 +108,6 @@ public func <- (left: inout GeoPoint?, right: (String, Map)) {
     left <- (map, transform)
 }
 
-/// Override operator used during the `propertyMapping(_:)` method.
-public func <- (left: inout GeoPoint!, right: (String, Map)) {
-    let (right, map) = right
-    let transform = GeoPointTransform()
-    kinveyMappingType(left: right, right: map.currentKey!, transform: transform)
-    left <- (map, transform)
-}
-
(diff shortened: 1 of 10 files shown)


Kinvey/Kinvey/Geolocation.swift
```diff
@@ -108,14 +108,6 @@ public func <- (left: inout GeoPoint?, right: (String, Map)) {
     left <- (map, transform)
 }
 
-/// Override operator used during the `propertyMapping(_:)` method.
-public func <- (left: inout GeoPoint!, right: (String, Map)) {
```
thomasconner: @heyzooi Can you explain a little bit what this function is for, why it was removed, and will it break anything?

Kinvey/KinveyApp/PerformanceTestData.swift
```diff
@@ -14,7 +14,7 @@ class PerformanceTestData: UIViewController {
     let client = Client(appKey: "kid_b1d6IY_x7l", appSecret: "079412ee99f4485d85e6e362fb987de8")
     
     func store<T: Persistable>() -> DataStore<T> where T: NSObject {
-        return DataStore<T>.collection(.network, client: self.client)
+        return DataStore<T>.collection(.network, options: Options(client: self.client))
```
thomasconner: Is this a sample app? Did a new warning appear due to an API change made sometime earlier to the SDK?

Kinvey/SSOApp/SSOApp1/AppDelegate.swift
```diff
@@ -51,7 +51,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, open url: URL, options: [UIApplicationOpenURLOptionsKey : Any] = [:]) -> Bool {
-        if User.login(redirectURI: micRedirectURI, micURL: url, authServiceId: authServiceId) {
+        if User.login(redirectURI: micRedirectURI, micURL: url, options: Options(authServiceId: authServiceId)) {
```
thomasconner: Is this a sample app? Did a new warning appear due to an API change made sometime earlier to the SDK?

Kinvey/Kinvey/Geolocation.swift
heyzooi: the compiler was warning/complaining the `Geopoint!` unnecessary overloads the `Geopoint?` version, which make sense, but running the unit tests the compiler is not smart enough to rewire the call for the optional version of the method. So I guess we will need to keep this warning for now since it seems to be a compiler bug. I will confirm that once again to make sure and if so I will create a ticket for the Swift compiler team.

Kinvey/KinveyApp/PerformanceTestData.swift
heyzooi: yes, this is not necessary related to Swift 4.1, but a deprecated usage for a while ago

Kinvey/SSOApp/SSOApp1/AppDelegate.swift
heyzooi: yeah, same situation

heyzooi:
I reverted some of changes that I made since there's a bug in the Swift compiler https://bugs.swift.org/browse/SR-6690

PR #299, opened 2018-06-18
MLIBZ-1937: returning support for tvOS for CocoaPods

heyzooi:
#### Description

Testing this again I actually figured the problem is actually a bug in CocoaPods that is solved after clear the CocoaPods cache folder like described here:

https://github.com/realm/realm-cocoa/issues/4758

#### Changes

- `Kinvey.podspec` file to support tvOS again for CocoaPods

#### Tests

- No need


heyzooi:
@vinaygahlawat i don't think that's necessary since it only happens in a very specific situation where CocoaPods cache seems to get corrupted, which makes this hard to reproduce in another machine. So don't think this will cause any impact in our support team.

PR #307, opened 2018-07-13
MLIBZ-2565: making p2p live stream internal instead of public

heyzooi:
#### Description

P2P Live Stream is not support any more

#### Changes

- All `LiveStream` symbols are now `internal` instead of `public`

#### Tests

- There's one unit test that had to be comment out


Kinvey/Kinvey/Realtime.swift
```diff
@@ -504,11 +504,11 @@ extension LiveStreamAclGroups: StaticMappable {
...
     
-    public var hashValue: Int {
+    internal var hashValue: Int {
         return uuid.hashValue
     }
     
-    public static func ==(lhs: LiveStream<Type>, rhs: LiveStream<Type>) -> Bool {
+    internal static func ==(lhs: LiveStream<Type>, rhs: LiveStream<Type>) -> Bool {
         return lhs.uuid == rhs.uuid
     }
```
vinaygahlawat: Since some of these deprecated methods are now just internal, should we just delete them?

heyzooi: sure!

Kinvey/KinveyTests/RealtimeAclTestCase.swift
```diff
@@ -11,9 +11,9 @@ import Kinvey
...
 class RealtimeAclTestCase: KinveyTestCase {
     
-    func testLiveStreamAclConsturctorNoParams() {
-        let acl = LiveStreamAcl()
-        XCTAssertNotNil(acl)
-    }
+//    func testLiveStreamAclConsturctorNoParams() {
+//        let acl = LiveStreamAcl()
+//        XCTAssertNotNil(acl)
+//    }
```
vinaygahlawat: Should we just delete this?

Kinvey/KinveyTests/RealtimeTestCase.swift
```diff
@@ -1,1750 +1,1750 @@
...
+//            user.registerForRealtime() {
+//                switch $0 {
+//                case .success:
+//                    registered = true
+//                    XCTAssertNotNil(user.realtimeRouter)
+//                case .failure:
+//                    XCTFail()
+//                }
+//
+//                expect
```
vinaygahlawat: Same here: should we just delete this instead of commenting out?

Kinvey/RealtimeSender/ViewController.swift
```diff
@@ -124,22 +124,22 @@ class ViewController: NSViewController {
...
+//                case .success:
+//                    print("Data Sent!")
+//                    exit()
+//                case .failure(let error):
+//                    print("Send Data Failed")
+//                    print(error)
+//                    exit()
+//                }
+//            }
         }
```
vinaygahlawat: Delete instead of commenting out?

PR #308, opened 2018-07-16
MLIBZ-2584: additional unit test for auto-pagination

heyzooi:
#### Description

Additional unit test for auto-pagination

#### Changes

- No changes

#### Tests

- Some additional checks for the unit test

### DIFF
--- a/Kinvey/KinveyTests/NetworkStoreTests.swift
+++ b/Kinvey/KinveyTests/NetworkStoreTests.swift
@@ -3477,20 +3477,17 @@ class NetworkStoreTests: StoreTestCase {
     }
     
     func testAutoPaginationEnabledNetworkOnly() {
-        var count = 0
         mockResponse { (request) -> HttpResponse in
-            defer {
-                count += 1
-            }
-            switch count {
-            case 0:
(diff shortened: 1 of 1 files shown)


Kinvey/KinveyTests/NetworkStoreTests.swift
```diff
@@ -3518,6 +3516,14 @@ class NetworkStoreTests: StoreTestCase {
             switch result {
             case .success(let products):
                 XCTAssertEqual(products.count, 21000)
+                XCTAssertNotNil(products.first)
+                if let product = products.first {
+                    XCTAssertEqual(product.entityId, "1")
+                }
+                XCTAssertNotNil(products.last)
+                if let product = products.last {
+                    XCTAssertEqual(product.entityId, "21000")
```
vinaygahlawat: Out of curiosity, why is the `entityId` value here `21000`, and not `21001`?

heyzooi: the range is starting from 0 to 20999, so that's why i'm doing `+ 1` in my mock response

PR #309, opened 2018-07-17
MLIBZ-2620: bugfix to persist Swift.Codable in cache

heyzooi:
#### Description

Bugfix to persist Swift.Codable in cache

#### Changes

- `Codable` entities cannot be persisted using the dynamic cache

#### Tests

- Added a unit test to test `pull()` a `find()` using a Sync DataStore


heyzooi:
@vinaygahlawat the user can call a `CustomEndpoint` and either the request body as well as the response body can be a custom type, which will be parsed from / to JSON

PR #313, opened 2018-07-31
MLIBZ-2636: replacing Realm ThreadSafeReference usage

heyzooi:
#### Description

Replacing Realm `ThreadSafeReference` usage since it can't be used inside a transaction

#### Changes

- Using the `primaryKey` as a reference instead of a `ThreadSafeReference` instance

#### Tests

- Unit test added to cover the use case


heyzooi:
@vinaygahlawat no, this is not part of the cache schema

PR #315, opened 2018-08-23
MLIBZ-2655: Cascade Delete References

heyzooi:
#### Description

Cascade Delete should not delete references that are being used by other objects

#### Changes

- Cascade Delete method

#### Tests

- A few unit tests added to cover cases where objects are deleted, but their references should not be deleted

### DIFF
--- a/Kinvey/Kinvey/RealmCache.swift
+++ b/Kinvey/Kinvey/RealmCache.swift
@@ -715,6 +715,31 @@ extension RealmSwift.NotificationToken: NotificationToken {
 
 extension RealmCache: DynamicCacheType {
     
+    private func countReferences(realm: Realm, objectClassName: String, nestedObject: Object, threshold: Int? = nil) -> Int {
+        var count = 0
+        for schema in realm.schema.objectSchema {
+            guard schema.className != objectClassName else {
+                continue
+            }
+            
+            for property in schema.properties {
+                guard property.objectClassName == objectClassName else {
+                    continue
+                }
+                
+                if property.isArray {
+                    count += realm.dynamicObjects(schema.className).filter("ANY \(property.name) == %@", nestedObject).count
+                } else {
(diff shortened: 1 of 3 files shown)


Kinvey/KinveyTests/SyncStoreTests.swift
```diff
@@ -7919,4 +7919,568 @@ class SyncStoreTests: StoreTestCase {
...
+        
+        if let person = items.last {
+            XCTAssertEqual(person.references.first?.entityId, referenceId)
+            let refreshedInstance = try! store.find(person.entityId!).waitForResult(timeout: defaultTimeout).value()
+            XCTAssertEqual(refreshedInstance.references.first?.entityId, referenceId)
+        }
+    }
+    
+    func testCascadeDeleteItemsWithOtherReferencesInAList() {
+        let referenceId = UUID().uuidString
```
vinaygahlawat: What is the difference between this function and the previous `func testCascadeDeleteItemsWithOtherReferencesInALists()`?

Kinvey/Kinvey/RealmCache.swift
```diff
@@ -715,6 +715,31 @@ extension RealmSwift.NotificationToken: NotificationToken {
...
+                }
+                
+                if property.isArray {
+                    count += realm.dynamicObjects(schema.className).filter("ANY \(property.name) == %@", nestedObject).count
+                } else {
+                    count += realm.dynamicObjects(schema.className).filter("\(property.name) == %@", nestedObject).count
+                }
+                if let threshold = threshold, count > threshold {
+                    return count
+                }
```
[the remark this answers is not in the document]
heyzooi: if a `threshold` is defined it means that we don't have to count for more items, so this function will run a little faster in this case

Kinvey/KinveyTests/SyncStoreTests.swift
heyzooi: the one above is a scenario where you have 2 lists, this one is a scenario where you have an object and a list

PR #317, opened 2018-08-27
MLIBZ-2637: refresh user should not update the refresh token

heyzooi:
#### Description

- Refresh User's method now checks if an existing `refresh_token` already exists and keep it as is

#### Changes

- `User.refresh()`

#### Tests

- Unit tests added to reproduce the issue


heyzooi:
do you mean `access_token` for facebook for example? that's actually a good question. let me do some research.

PR #318, opened 2018-09-04
MLIBZ-2661: bugfix for large pull memory consumption

heyzooi:
#### Description

Realm have a limitation of keeping all objects in memory during a single large transaction. To avoid this situation we did a few changes around the cache layer.

#### Changes

- Request Start persisted is now from the `count()` 

#### Tests

- A few unit tests to ensure the cache data still consistent

### DIFF
--- a/Kinvey/KinveyTests/NetworkStoreTests.swift
+++ b/Kinvey/KinveyTests/NetworkStoreTests.swift
@@ -771,6 +771,48 @@ class NetworkStoreTests: StoreTestCase {
         }
     }
     
+    func testTranslateQueryCodable() {
+        let store = try! DataStore<PersonCodable>.collection(.sync)
+        let id = UUID().uuidString
+        
+        if useMockData {
+            mockResponse { (request) -> HttpResponse in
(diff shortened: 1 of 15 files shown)


ivanstoyanov:
Would this limitation also apply to Android? Can we make the same test there?

Kinvey/KinveyTests/NetworkStoreTests.swift
```diff
@@ -771,6 +771,48 @@ class NetworkStoreTests: StoreTestCase {
         }
     }
     
+    func testTranslateQueryCodable() {
+        let store = try! DataStore<PersonCodable>.collection(.sync)
```
vinaygahlawat: Why is a `(.sync)` store type in the `NetworkStoreTests` file?

heyzooi: good call, let me move that do the `SyncStoreTests`

heyzooi:
@ivanstoyanov for sure worth to test, but I would guess it also happen in Android since the code base is the same.

PR #319, opened 2018-09-10
Codable for EmailVerification and PasswordReset metadata objects

ifabijanovic:
#### Description
`EmailVerification` and `PasswordReset` metadata objects which are a part of `UserMetadata` do not have the `Codable` protocol implemented. This caused the `User` object to fail to initialize using `init(from:)` with the following error:

screen shot 2018-09-10 at 10 50 05

#### Changes
Implemented `Codable` for both `EmailVerification` and `PasswordReset` objects. Also updated the `mapping(map:)` method of `Metadata` as it seems it wasn't passing the correct keys to `map`.

#### Tests
Tested in an app making use of Kinvey user accounts, the above mentioned error is gone, checked that all of the data was correctly parsed and available inside the `User` instance.

### DIFF
--- a/Kinvey/Kinvey/Metadata.swift
+++ b/Kinvey/Kinvey/Metadata.swift
@@ -130,9 +130,9 @@ public class Metadata: Object, Codable {
     /// This function is where all variable mappings should occur. It is executed by Mapper during the mapping (serialization and deserialization) process.
     @available(*, deprecated: 3.18.0, message: "Please use Swift.Codable instead")
     public func mapping(map: Map) {
-        lmt <- (CodingKeys.lastModifiedTime.rawValue, map[CodingKeys.lastModifiedTime.rawValue])
-        ect <- (CodingKeys.entityCreationTime.rawValue, map[CodingKeys.entityCreationTime.rawValue])
-        authtoken <- (CodingKeys.authtoken.rawValue, map[CodingKeys.authtoken.rawValue])
+        lmt <- (CodingKeys.lastModifiedTime.rawValue, map[CodingKeys.lastModifiedTime])
+        ect <- (CodingKeys.entityCreationTime.rawValue, map[CodingKeys.entityCreationTime])
+        authtoken <- (CodingKeys.authtoken.rawValue, map[CodingKeys.authtoken])
     }
 
 }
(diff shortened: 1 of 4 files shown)


heyzooi:
Hi @ifabijanovic, thanks for your contribution.
We need to add unit tests as well. Maybe we can do that after we accept this PR. Let me check the best way to do this.

heyzooi:
@ifabijanovic I'm wondering if would be possible for you to add a unit test, even a simple one and we could add more later if needed.

ifabijanovic:
I've implemented `Codable` for `MyUser` in `UserTests.swift` but this has an effect of tests checking `socialIdentity` to start failing. I am also getting a SIGABRT when running just UserTests in test `testMICCancelUserAction` on line 4835.

heyzooi:
I suggest you create your own user class (a new one, like `MyUserCodable`) so it will not affect the existing unit tests.

ifabijanovic:
That does not make sense - you want to make use of existing unit tests when introducing changes and this change found that you have a bug in `UserSocialIdentity` not implementing `Codable`. I am also not sure why `testMICCancelUserAction` is timing out when just running `UserTests` because it passes when running ALL tests in `KinveyTests` target - unit tests are not atomic?

heyzooi:
@ifabijanovic I can't reproduce the issue that you are reporting with the `testMICCancelUserAction` test. Running it alone or all tests of `UserTests`, I'm always getting a success result.

I suggest you create your own user class and duplicate the unit tests to cover the `Codable` use case.

ifabijanovic:
That's strange that everything passes for you... Anyway I've reverted my original unit test changes and added a new custom user class as you advised. I implemented two unit tests (based off other unit tests) - save one works, but refresh one fails for the custom property. It seems that updating a custom user property on the client side results in correct behaviour, but updating a custom value on the backend is not propagated to the client after calling `refresh`.

ifabijanovic:
Nice changes! I think `User` should be good now - I will give this a try in my project. We don't use any 3rd party auth (Facebook, Google) so that part might still need some work in regards to `Codable` - the `socialIdentities` bit does not implement it yet.

PR #329, opened 2018-10-17
MLIBZ-2714: Improvement for observe objects

heyzooi:
#### Description

Improvement for observe objects

#### Changes

- A update is now performed instead of a delete and insert for existing objects

#### Tests

- Same unit tests since it was more like an improvement than a fix for the `update` vs `delete/insert`


vinaygahlawat:
@heyzooi - Is this done in all update cases for Realm? If so, what is the performance cost of this change?

heyzooi:
@vinaygahlawat I noticed a little gap in memory consumption after update to the newest Realm version. It was from 60mb to something around 64mb.

heyzooi:
@vinaygahlawat Realm have this limitation the primary key cannot be changed. So for temporary objects we need to delete and insert the new one (with the new _id)

https://realm.io/docs/swift/latest#primary-keys

PR #330, opened 2018-10-22
MLIBZ-2719: Prevent multiple calls to refresh token

heyzooi:
#### Description

Prevent multiple calls to refresh token which was causing a race condition

#### Changes

- `User`, `Client` and `HttpRequest` operations are queued when a `401` response is returned

#### Tests

- Unit test added to reproduce a multiple calls being made when the token needs to be refreshed


Kinvey/Kinvey/HttpRequest.swift
```diff
@@ -405,30 +410,49 @@ internal class HttpRequest<Result>: TaskProgressRequest, Request {
...
-                            self.credential = user
-                            self.execute(urlSession: urlSession, completionHandler)
-                        case .failure(let error):
-                            if let error = error as? Kinvey.Error {
-                                switch error {
-                                case .invalidCredentials:
-                                    if let user = self.credential as? User {
-                                        user.logout()
+                    DispatchQueue.global(qos: .default).async {
+                        guard !client.refreshingToken else {
```
vinaygahlawat: Is the `client.refreshingToken` made thread-safe because it is being accessed and set from within the `DispatchQueue`?

heyzooi: right, so this is a serial queue and I'm calling `wait()` on the `DispatchGroup` so the combination of both makes the entire refresh token process thread-safe

PR #335, opened 2018-11-02
MLIBZ-2725: Be able to store dates as ISO 8601 strings

heyzooi:
#### Description

Be able to store dates as ISO 8601 strings

#### Changes

- Users using `Codable` don't have an easy way to store dates as before

#### Tests

- No need for new unit tests

### DIFF
--- a/Kinvey/Kinvey/Date.swift
+++ b/Kinvey/Kinvey/Date.swift
@@ -14,4 +14,8 @@ extension Date {
         return KinveyDateTransform().transformToJSON(self)!
     }
     
+    public func toISO8601() -> String {
+        return toString()
+    }
+    
 }
--- a/Kinvey/Kinvey/Date.swift
+++ b/Kinvey/Kinvey/Date.swift
@@ -9,13 +9,9 @@
 import Foundation
 
 extension Date {
-    
-    func toString() -> String {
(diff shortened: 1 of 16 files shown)


Kinvey/Kinvey/Date.swift
```diff
@@ -9,8 +9,8 @@
 import Foundation
 
 extension Date {
-    
-    func toString() -> String {
```
vinaygahlawat: Doesn't appear to have been public, but is it possible that current developers can call `toString()` on a `Date` object? Does the `toString()` method have to be deprecated?

heyzooi: oh no need since the method was internal. There's no way to call an internal method.

heyzooi: I think we can use the `ISO8601DateFormatter` now since we bumped the minimum version on all platforms. Doing that now.

heyzooi: it will need to wait until next year since the `milliseconds` is only available in the next minimum bump version, so we need to keep as is for now

https://developer.apple.com/documentation/foundation/iso8601dateformatter/options/2923300-withfractionalseconds

PR #337, opened 2018-11-14
MLIBZ-1085: Adding new store type .auto

heyzooi:
#### Description

Adding a new store type that first hits the network (on read operations). If the network request fails we return whatever data is available in the local cache.
For write operations we can use the existing `WritePolicy.localThenNetwork`

#### Changes

- Deprecating the existing `.cache` store type and implementing the new `.auto` store type

#### Tests

- New unit test added to reproduce a scenario using the new `.auto` store type


Kinvey/Kinvey/Tests.xcconfig
```diff
@@ -0,0 +1,12 @@
...
+//  Kinvey
+//
+//  Created by Victor Hugo Carvalho Barros on 2018-11-14.
+//  Copyright © 2018 Kinvey. All rights reserved.
+//
+
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
+
+SWIFT_SUPPRESS_WARNINGS = YES
```
vinaygahlawat: What is this file for?

Kinvey/Kinvey/ReadOperation.swift
```diff
@@ -63,6 +63,18 @@ extension ReadOperationType {
...
             }
             return AnyRequest(request)
+        case .networkOtherwiseLocal:
+            let requests = MultiRequest<Result<SuccessType, FailureType>>()
+            let networkRequest = executeNetwork() {
+                switch $0 {
+                case .success:
+                    completionHandler?($0)
+                case .failure:
+                    requests.addRequest(self.executeLocal(completionHandler))
```
vinaygahlawat: Need to check why we failed here because we only want to fetch from the local datastore if a network connectivity issue occurred. If a 4xx/5xx response comes back, we still need to correctly handle this exception by alerting the user.

Kinvey/Kinvey/ReadPolicy.swift
```diff
@@ -20,4 +20,7 @@ public enum ReadPolicy: UInt {
     /// Read first from the local cache and then try to get data from the network (backend).
     case both
     
+    /// Tries to get the most recent data from the backend first, if it fails it returns data from the local cache.
```
vinaygahlawat: Only should return data from the local cache if the server cannot be reached (i.e. does not return a result we can handle - 4xx/5xx).

Kinvey/Kinvey/FileStore.swift
```diff
@@ -824,7 +824,7 @@ open class FileStore<FileType: File> {
     @discardableResult
     fileprivate func downloadFileURL(
         _ file: FileType,
-        storeType: StoreType = .cache,
+        storeType: StoreType,
```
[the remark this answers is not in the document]
heyzooi: good question. Also this should be a deprecation and add a new method without the default value

Kinvey/Kinvey/Tests.xcconfig
heyzooi: this is a configuration file for the unit tests, since we do test some deprecated methods I added this to remove the warnings for the test bundle

Kinvey/Kinvey/FileStore.swift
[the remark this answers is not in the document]
heyzooi: sorry this is a private function, so it's correct

PR #345, opened 2019-01-04
SwiftLint warning fix

heyzooi:
#### Description

New version of SwiftLint detected a new warning

#### Changes

- Minor change how the `hashValue` is calculated. `hashValue` is deprecated and the new `Hasher` should be used instead

#### Tests

- Same unit tests


heyzooi:
@vinaygahlawat the swift runtime is smart enough to redirect the call to the new way if the class implements the new method. To make sure I just added a check in one of our existing unit tests.

PR #346, opened 2019-01-04
MLIBZ-2838: Bugfix crash on signpost calls

heyzooi:
#### Description

Adding {public} to the signpost calls when handled strings

#### Changes

- Signpost calls now uses `%@` instead of `%s` for string formatting

#### Tests

- Same unit tests


heyzooi:
@vinaygahlawat sorry, just fixed, `%@` is the correct one

PR #349, opened 2019-02-01
MLIBZ-2870: Converting dates in queries to ISO 8601

heyzooi:
#### Description

Converting dates in queries to ISO 8601 so queries will return the expected values in both local and network calls

#### Changes

- JSON Decoder and Encoder now converts ISO 8601 by default
- Allow users to set custom JSON Decoder and Encoder
- Dates on queries converts into ISO 8601 but only for network calls

#### Tests

- Tests for query local and network calls

### DIFF
--- a/Kinvey/Kinvey/Date.swift
+++ b/Kinvey/Kinvey/Date.swift
@@ -9,9 +9,17 @@
 import Foundation
 
 extension Date {
+    
+    static let dateWriteFormatter: DateFormatter = {
+        let wFormatter = DateFormatter()
+        wFormatter.locale = Locale(identifier: "en_US_POSIX")
+        wFormatter.timeZone = TimeZone(identifier: "UTC")
+        wFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+        return wFormatter
+    }()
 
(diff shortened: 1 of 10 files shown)


Kinvey/Kinvey/Date.swift
```diff
@@ -9,9 +9,17 @@
...
+        let wFormatter = DateFormatter()
+        wFormatter.locale = Locale(identifier: "en_US_POSIX")
+        wFormatter.timeZone = TimeZone(identifier: "UTC")
+        wFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+        return wFormatter
+    }()
 
     public func toISO8601() -> String {
-        return KinveyDateTransform().transformToJSON(self)!
+        return Date.dateWriteFormatter.string(from: self)
```
vinaygahlawat: @heyzooi - Because we are changing the format of what we are returning here and this is a `public func`, do we have to worry about this being a breaking change?

heyzooi: the format output is the same, I just made some changes here for performance reasons, so instead create a new instance and call a method every time, I'm using the static variable so we will have a better performance

PR #351, opened 2019-02-20
MLIBZ-2883: Delta Set in the DataStore not working

heyzooi:
#### Description

Change the value for Delta Set in the DataStore not working as it should

#### Changes

- We were ignoring the `newValue` and always set `true`

#### Tests

- TBD


heyzooi:
I checked and noticed the method was not used anymore and since it's a private method I decided to removed the `set` method for this property

PR #354, opened 2019-04-08
Changes required to compile and run all unit tests using Xcode 10.2

heyzooi:
#### Description

Changes required to compile and run all unit tests using Xcode 10.2

#### Changes

- Compiler errors fixed
- Race conditions detected
- Unit tests adjusments

#### Tests

- Same Unit Tests


heyzooi:
@vinaygahlawat i'm in touch with the Swift team to understand how we can specify the version number in deprecated symbols

https://bugs.swift.org/browse/SR-6356

heyzooi:
The Swift team replied saying the deprecated version does not work for frameworks, only for platforms and the advice was to add the version number to the deprecated message.

PR #363, opened 2019-06-23
MLIBZ-3060: Multi-Insert Push implementation

heyzooi:


Kinvey/Kinvey/Geolocation.swift
```diff
@@ -14,6 +14,8 @@ import MapKit
 /// Class that represents a 2D geolocation with latitude and longitude
 public final class GeoPoint: Object {
     
+    public static let CodingKey = "_geoloc"
```
vinaygahlawat: What is `_geoloc` used for in the multi-insert implementation?

Kinvey/Kinvey/HttpRequestFactory.swift
```diff
@@ -426,21 +426,22 @@ struct HttpAppDataRequestFactory: AppDataRequestFactory {
...
+        let isNewObj = isNew ?? Kinvey.isNew(entityId: objId)
         let request = HttpRequest<Result>(
             httpMethod: isNewObj ? .post : .put,
             endpoint: isNewObj ? AppDataEndpoint.appData(client: client, collectionName: collectionName) : AppDataEndpoint.appDataById(client: client, collectionName: collectionName, id: objId!),
             credential: client.activeUser,
             options: options
         )
         
-        if isNewObj {
+        if isNewObj || (Kinvey.isNew(entityId: objId)) {
```
vinaygahlawat: If we are defining `isNewObj` above to set it to `Kinvey.isNew(...)` if it is initially null, what is the reason for checking it here again?

Kinvey/Kinvey/Kinvey.swift
```diff
@@ -327,9 +327,13 @@ func buildError(
         let debug = json["debug"],
         let description = json["description"]
     {
-        let refreshToken = client.activeUser?.socialIdentity?.kinvey?["refresh_token"] as? String
-        if refreshToken == nil {
-            client.activeUser?.logout()
+        var logoutOnDefer = true
```
vinaygahlawat: How is the checking of `refresh_token` related to multi-insert push functionality?

Kinvey/Kinvey/Geolocation.swift
heyzooi: some tests are related to send invalid geolocations, like (-300, -300)
since this is a special field handled in the backend I decided to create this constant so we (and also developers) don't have to repeat `_geoloc` everywhere

Kinvey/Kinvey/Kinvey.swift
heyzooi: we were not handling the `InsufficientCredentials` error before, which should not logout the user.
a scenario detected by the unit tests

Kinvey/Kinvey/HttpRequestFactory.swift
heyzooi: you are right, removing it now!

heyzooi: thanks for notice this

PR #370, opened 2019-09-06
MLIBZ-3153: Removing any mention of UIWebView

heyzooi:
#### Description

Since Apple is now complaining about the code having any mention about `UIWebView` during the App Review process (even if you are not using `UIWebView`), we are removing all mentions of `UIWebView`

#### Changes

- Deprecating the usage of `MICUserInterface.uiWebView` in favor of `MICUserInterface.wkWebView`
- Removing all `UIWebView` mentions

#### Tests

- Had to remove some `UIWebView` related classes


Kinvey/Kinvey/MIC.swift
```diff
@@ -361,7 +362,7 @@ public enum MICApiVersion: String {
 import UIKit
 import WebKit
 
-class MICLoginViewController: UIViewController, WKNavigationDelegate, UIWebViewDelegate {
+class MICLoginViewController: UIViewController {
```
vinaygahlawat: Is this only used internally, or could this be a breaking change for developers?

Kinvey/KinveyTests/UserTests.swift
```diff
@@ -5122,115 +5122,115 @@ extension UserTests {
...
-            }
-            
-            expectationLogin?.fulfill()
-        }
-        
-        waitForExpectations(timeout: defaultTimeout) { error in
-            expectationLogin = nil
-        }
-    }
+//    func testMICTimeoutAction() {
```
vinaygahlawat: Is there a good reason for keeping this commented test, or should we delete it completely?

Kinvey/Kinvey/MIC.swift
heyzooi: that's an internal class, so, no breaking changes at this point, but I end up deciding to remove UIWebView completely because of the tests were not working, meaning, keeping the `MICUserInterface.uiWebView` could cause bugs so, we will have to make a major release `4.0`

Kinvey/KinveyTests/UserTests.swift
heyzooi: decided to remove UIWebView completely and all tests related, which will cause a breaking change, so a major release will be necessary

PR #384, opened 2020-10-30
KDEV-1151: Respect maximum items per request limit on push

mbektchiev:
#### Description
When the number of entities in a `save()` or `push()` request exceeds the max allowed in the request (100 entities), the entities should be broken down into multiple sets of the max amount allowed, and multiple multi-insert requests should be sent out. `push()` however didn't follow this logic.

refs https://kinvey.atlassian.net/browse/KDEV-1151

#### Changes
* Split into multiple `pendingOperations` when items to
sync are more than `maxSizePerRequest` (100)
* Remove `;` which are reported as linter warnings
* Add `.swiftpm` to `.gitignore`

#### Tests
* Test with the sample provided in the Jira item
* Add a test case for push with a number of items which is
1,1 times greater than the limit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ Kinvey.framework.zip
 scripts/RealtimeSend/Rome
 Package.resolved
 Kinvey.xcodeproj
+.swiftpm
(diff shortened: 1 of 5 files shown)


alexanderfilipov (approved):
LGTM.

PR #385, opened 2021-09-24
KDEV-1928: Support Xcode 13

ventsislav-georgiev:
Changes:
* Update all dependencies to latest versions
  * Most notably Realm from v3.21.0 to v10.15.1
* Build dependencies as xcframeworks (this is required for the new build system of Xcode)
  * With latest Carthage (0.38) build with --use-xcframeworks
  * Cleanup all Copy Carthage frameworks build phase tasks and add the test frameworks where appropriate
  * Update all framework search paths 
* Switch testing for iOS to target iOS 15.0
  * The push notification device register test is skipped as it pops a system alert which was previously confirmed with KIF, but as of iOS 14 it doesn't support acknowledging system alerts. A manual local run of the test can be confirmed passing by targeting iOS 13.
> The removeAllMemoryLeak test was failing after the update. After diagnostic runs, the Xcode tooling did not report any memory leaks and also the memory reported from the tests did not match the one from Xcode's own profiling. It seemed that the memory reported from the test was the Real Memory of the process but Xcode was showing the Private Memory. We changed the method used for getting the currently used memory in the tests to fix the issue and taking only the Private Memory into account.
* Switch development team used for codesigning
* Remove tests related dependencies from the Package.swift
* Switch to Github actions for CI
  * Cleanup cirrusci and travis configurations
  * New github workflow triggered when branches are targeting master and via cron on a daily basis
  * The dependency builds are really slow ~1h40m, so a cache is utilized for them based on the hash of `Cartfile.resolved` content.
  * The same cache is utilized for the Swift package build, it will not be triggered again unless the Package.swift file content changes. The Package.swift file was updated to use the exact version of dependencies matching the ones from `Cartfile.resolved`
 > The iOS tests are currently failing with compilation issues on Xcode 13 Github Action. Those can be fixed via the setting "Other linker flags" config to `-lstdlib++ -lz`. However, even then the tests fail to complete with an error: `Failed to establish communication with the test runner. (Underlying Error: The operation couldn’t be completed. (DTXProxyChannel error 1.))`. So for now those tests are running on macOS 10.15 with Xcode 12.4.


mbektchiev (approved):

mbektchiev (approved):
Looks good! :)

